PR #2, opened 2015-01-22
Sync mode for running independently of PaaS to register nodes

rajatchopra:
Do not merge. Needs testing.

@mrunalp review please.


main.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"path"
```
mrunalp: Needs gofmt.

PR #5, opened 2015-01-27
non-permanence for linux bridge, so that we do not need restart network service

rajatchopra:
@mrunalp review please

### DIFF
--- a/ovs-simple/bin/openshift-sdn-simple-setup-node.sh
+++ b/ovs-simple/bin/openshift-sdn-simple-setup-node.sh
@@ -4,9 +4,9 @@ set -ex
 
 echo $@
 
+## openvswitch
 ovs-vsctl del-br br0 || true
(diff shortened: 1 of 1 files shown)


ovs-simple/bin/openshift-sdn-simple-setup-node.sh
```diff
@@ -19,23 +19,24 @@ ip link set vovsbr txqueuelen 0
...
-NETMASK=255.255.255.0
-STP=yes
-EOF
-service network restart || true
+
+## linux bridge
+ip link set lbr0 down || true
+brctl delbr lbr0 || true
+brctl addbr lbr0
+ip addr add ${1}/24 dev lbr0
```
mrunalp: Need to remove this hard code at some point.

mrunalp:
LGTM

rajatchopra:
Yeah, that hardcode is just the replacement for Netmask=255.255.255.0 that existed before. It should infact be easier now.

PR #7, opened 2015-01-30
Use $DOCKER_OPTIONS env var for docker settings if present

dobbymoodge:
Use `$DOCKER_OPTIONS` env var for docker settings if present

### DIFF
--- a/ovs-simple/bin/openshift-sdn-simple-setup-node.sh
+++ b/ovs-simple/bin/openshift-sdn-simple-setup-node.sh
@@ -48,8 +48,9 @@ if ! grep -q "^OPTIONS='${DOCKER_OPTIONS}'" /etc/sysconfig/docker
 then
     cat <<EOF > /etc/sysconfig/docker
 # This file has been modified by openshift-sdn. Please modify the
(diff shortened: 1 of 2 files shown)


ovs-simple/bin/openshift-sdn-simple-setup-node.sh
```diff
@@ -39,7 +39,21 @@ iptables -I INPUT $lineno -p udp -m multiport --dports 4789 -m comment --comment
 
 
 ## docker
-echo "OPTIONS='-b=lbr0 --mtu=1450 --selinux-enabled'" >/etc/sysconfig/docker
+if [[ -z "${DOCKER_OPTIONS}" ]]
```
mrunalp: Could we document in the README or print a warning here when someone has set this variable that they better know what they are doing else openshift-sdn won't work correctly?

dobbymoodge: Seems fair. I'll put it in the README

mrunalp:
Talked with @rajatchopra and we are okay with this. LGTM.

Issue #12, opened 2015-02-10
openshift-sdn gives up too early or wrong return code


goldmann:
Sometimes when OpenShift is not running, or booting, SDN fails to connect to it and gives up, exiting with code `0`.

```
lut 10 13:53:57 openshift-master systemd[1]: Starting openshift SDN master...
lut 10 13:53:57 openshift-master systemd[1]: Started openshift SDN master.
lut 10 13:53:57 openshift-master openshift-sdn[1250]: I0210 13:53:57.135515 01250 main.go:108] Installing signal handlers
lut 10 13:53:57 openshift-master openshift-sdn[1250]: I0210 13:53:57.138897 01250 controller.go:47] Self IP : 127.0.0.1
lut 10 13:53:59 openshift-master openshift-sdn[1250]: E0210 13:53:59.141011 01250 controller.go:73] Error in initializing/fetching subnets - 501: All the given peers are not reachable (Tried to connect to each peer twice and failed) [0
```

It should try to reconnect automatically with no limit set on the retries or at least least exit with status != `0`, so systemd could restart the service.

goldmann:
As a workaround I use currently this:

```
Restart=always
RestartSec=5
```

In the systemd service definition, which is not really nice :)

rajatchopra:
Fixed with #14 , master should be more accommodating now. Please re-open if there are problems still.

PR #14, opened 2015-02-11
fix issue#12. Give some slack for etcd to come alive.

rajatchopra:
@mrunalp 
Review please.

### DIFF
--- a/main.go
+++ b/main.go
@@ -118,7 +118,7 @@ func main() {
 	} else if opts.master {
 		err := be.StartMaster(opts.sync)
 		if err != nil {
-			log.Errorf(err.Error())
+			log.Error(err)
 			return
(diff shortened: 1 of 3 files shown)


rajatchopra:
Fixes #12

main.go
```diff
@@ -118,6 +118,7 @@ func main() {
 	} else if opts.master {
 		err := be.StartMaster(opts.sync)
 		if err != nil {
+			log.Errorf(err.Error())
```
mrunalp: I think you can just use err here.

rajatchopra: Fixed. Errorf requires a string, but we do not use any formatting here.

mrunalp:
LGTM

PR #23, opened 2015-02-19
openshift-sdn does not require openvswitch or bridge-utils

sdodson:
Only openshift-sdn-node requires the overlay networking components.


mrunalp:
LGTM

PR #24, opened 2015-02-19
Always treat -etcd-path as a directory

Miciah:
`newSubnetRegistry()`: Pass `opts.etcdPath` and the subdirectory name (either "subnets" or "minions") to `path.Join` as separate arguments rather than catenating them as strings and passing the catenated string to `path.Join`.

Before this change, if `-etcd-path` were passed a pathname "/foo" that did not end with a slash, openshift-sdn would use directory names "/foosubnets" and "/foominions" instead of "/foo/subnets" and "/foo/minions".

### DIFF
--- a/main.go
+++ b/main.go
@@ -67,10 +67,10 @@ func newNetworkManager() (controller.Controller, error) {
 func newSubnetRegistry() registry.SubnetRegistry {
 	peers := strings.Split(opts.etcdEndpoints, ",")
 
-	subnetPath := path.Join(opts.etcdPath + "subnets")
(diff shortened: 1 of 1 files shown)


Miciah:
@rajatchopra, does this look good?

rajatchopra:
Yes. Looks good. Thanks for the fix. Trying to get a travis setup done, would probably use this as the first one to go through the grind.
Waiting for that, and will get this merged.

Miciah:
OK, understood.  Thanks for reviewing!

mrunalp:
LGTM. Let us get this one in for now. We need something like the [test] tag. I will work that out with dmcphers.

PR #28, opened 2015-02-23
WatchMinions: Correctly handle errors

Miciah:
If we get a `nil` response or a non-`nil` error, then retry the watch. Before this change, in the case of a `nil` response and a non-`nil` error, `WatchMinions` would try to dereference the `nil` response.


mrunalp:
@rajatchopra ping

PR #33, opened 2015-03-01
Fixed error handling to catch error from newNetworkManager

nak3:
Should catch this error https://github.com/openshift/openshift-sdn/blob/master/main.go#L116-L119

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,7 +62,7 @@ func newNetworkManager() (controller.Controller, error) {
 		host = strings.TrimSpace(string(output))
 	}
 
-	return controller.NewController(sub, string(host), opts.ip), nil
+	return controller.NewController(sub, string(host), opts.ip)
 }
 
 func newSubnetRegistry() registry.SubnetRegistry {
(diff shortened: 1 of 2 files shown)


nak3:
Reproduce the issue by following steps
----
1) Check current /etc/hostname

  `# cat /etc/hostname`

2) Chnange /etc/hostname to invalid name (which you cannot lookup LookupHost)

  `# echo "foo" > /etc/hostname`

3) Run openshift-sdn

  `./_output/local/go/bin/openshift-sdn`

4) You can see the runtime panic message

    ... snip ...

    E0303 10:09:40.042921 13500 controller.go:43] Failed to lookup IP Address for foo
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal 0xb code=0x1 addr=0x30 pc=0x401c63]

    goroutine 16 [running]:
    runtime.panic(0x6bd3c0, 0x88e873)
     	/usr/lib/golang/src/pkg/runtime/panic.c:279 +0xf5
    main.main()
     	/root/openshift-sdn/_output/local/go/src/github.com/openshift/openshift-sdn/main.go:126 +0x633

    ... snip ...

rajatchopra:
Looks good. Thanks.

PR #34, opened 2015-03-02
Remove unnecessary print

nak3:
I believe these fmt.Println are not necessary. If they are necessary, they should use log.Info or other formal message.


mrunalp:
LGTM

PR #35, opened 2015-03-02
Add Restart=on-failure to unit files

Miciah:
Add `Restart=on-failure` to the unit files for openshift-sdn-master and openshift-sdn-node.

### DIFF
--- a/rel-eng/openshift-sdn-master.service
+++ b/rel-eng/openshift-sdn-master.service
@@ -8,6 +8,8 @@ Requires=openshift-master.service
 Type=simple
 EnvironmentFile=-/etc/sysconfig/openshift-sdn-master
 ExecStart=/usr/bin/openshift-sdn $OPTIONS
+Restart=on-failure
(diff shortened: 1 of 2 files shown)


Miciah:
@sdodson, :+1:?

sdodson:
Perhaps set RestartSec=1s ? By default it's going to sleep only 100ms, if we're trying to guard against openshift-master startup time that may not be long enough. Otherwise :+1:

sdodson:
Also, FWIW, I was looking at adding systemd notify support to openshift. That would allow openshift to signal to systemd when it had finished starting up so that systemd would wait to start dependencies until after that had happened. However, given that these services aren't necessarily located on the same system we should still restart as you're proposing.

Miciah:
Updated.

sdodson:
:+1:

mrunalp:
@Miciah @sdodson thanks.
LGTM

PR #36, opened 2015-03-02
Improve output formatting

Miciah:
Use clearer and more consistent formatting for output.

Consistently format error values using `%v` with the error values themselves rather than using `%s` and `err.Error()`.

Indicate that we are printing the container network, local host subnet, and gateway for the latter during initialization of openshift-sdn in node mode rather than cryptically printing IP addresses and CIDR suffixes with no explanation of what they are.

Indicate that we are printing the subnet prefix length during initialization of openshift-sdn in master mode rather than cryptically printing a lone number with no explanation of what it means.

### DIFF
--- a/ovs-simple/bin/openshift-sdn-simple-setup-node.sh
+++ b/ovs-simple/bin/openshift-sdn-simple-setup-node.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-echo $@
+printf 'Container network is "%s"; local host has subnet "%s" and gateway "%s".\n' "$3" "$2" "$1"
 
 ## openvswitch
 ovs-vsctl del-br br0 || true
--- a/pkg/netutils/subnet_allocator.go
+++ b/pkg/netutils/subnet_allocator.go
@@ -19,7 +19,7 @@ func NewSubnetAllocator(network string, capacity uint, inUse []string) (*SubnetA
 
 	netMaskSize, _ := netIP.Mask.Size()
 	if capacity > (32 - uint(netMaskSize)) {
-		return nil, fmt.Errorf("Subnet capacity cannot be larger than number of networks available")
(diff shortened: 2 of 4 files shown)


Miciah:
@rajatchopra, do these look reasonable?

mrunalp:
LGTM

rajatchopra:
@Miciah rebase please. Looks ready otherwise.
Thanks.

pkg/netutils/subnet_allocator.go
```diff
@@ -18,8 +18,9 @@ func NewSubnetAllocator(network string, capacity uint, inUse []string) (*SubnetA
 	}
 
 	netMaskSize, _ := netIP.Mask.Size()
+	fmt.Printf("Subnet prefix length is %d bits.\n", netMaskSize)
```
Miciah: Before I rebased this PR, there had been a `fmt.Println(netMaskSize)` that this commit deleted before adding the more explicit `fmt.Printf` statement.  I can delete this `fmt.Printf` if you believe it is superfluous.

ovs-simple/bin/openshift-sdn-simple-setup-node.sh
```diff
@@ -2,7 +2,7 @@
 
 set -ex
 
-echo $@
+printf 'Container network is "%s"; local host has subnet "%s" and gateway "%s".\n' "$3" "$2" "$1"
```
Miciah: This will conflict with #38; I can rebase again once it merges.

pkg/netutils/subnet_allocator.go
mrunalp: We can remove this line. It was a debugging line I forgot to remove.

rajatchopra:
We can rebase #38, this is ready now.

PR #37, opened 2015-03-02
Elaborate on how openshift-sdn functions

Miciah:
Add more details on how openshift-sdn works to the readme.

Note that in the future, containers will be connected directly to OVS, and lbr0 will go away, so the description that this commit adds will need to be amended accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,33 @@
-## SDN solutions for Openshift
+## SDN solution for OpenShift
 
-Software to get an overlay network up and running for a docker cluster. This is still a work in progress. Do not use it in production.
+openshift-sdn configures an overlay network for a Docker cluster, using Open
+vSwitch (OVS). This is still a work in progress. Do not use it in production.
+
+The openshift-sdn daemon runs in one of two modes: either master or minion.
+
+If launched without the -minion flag, openshift-sdn operates as a master.  In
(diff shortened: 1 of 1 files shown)


sdodson:
This is very helpful.

README.md
```diff
@@ -59,6 +86,54 @@ Back on the master, to finally register the node:
...
+eth0 (in A's netns) -> vethA -> lbr0 -> vlinuxbr -> vovsbr -> br0 -> vovsbr ->
+vlinuxbr -> lbr0 -> vethB -> eth0 (in B's netns).
+
+Next suppose instead that container A is on the local host and container B is on
+a remote host on the same container network.  Then the flow of packets from
+container A to container B is as follows:
+
+eth0 (in A's netns) -> vethA -> lbr0 -> vlinuxbr -> vovsbr -> br0 -> vxlan0 ->
+network\* -> vxlan0 -> br0 -> vovsbr -> vlinuxbr -> lbr0 -> vethB -> eth0 (in
+B's netns).
```
rajatchopra: Can we move this section to another document? This flow is true of the somewhat un-optimized version in master today, but the underlying architecture will change soon. It will be easier to maintain, especially when we start giving other solutions as choice to the user.

Miciah:
Closing in favour of openshift/openshift-docs#491.

PR #44, opened 2015-03-23
Make service dependency ordering more strict

sdodson:
This formalizes service startup ordering.
Before/After only affect ordering.
Wants/Requires affects dependencies, wants being a weaker form of Requires that will trigger the dependency to start but will carry on if the dependency fails.

My only concern is that I wonder if we should add Before=docker.service because if we don't and we've run openshift-sdn previously it will have configured docker to use lbr0 which will not exist after a reboot before openshift-sdn has created it.

### DIFF
--- a/rel-eng/openshift-sdn-node.service
+++ b/rel-eng/openshift-sdn-node.service
@@ -1,8 +1,13 @@
 [Unit]
 Description=OpenShift SDN Node
-Before=openshift-node.service
 After=openvswitch.service
+After=openshift-sdn-master.service
+After=openshift-master.service
+Wants=openshit-sdn-master.service
+Wants=openshift-master.service
 Requires=openvswitch.service
+Before=openshift-node.service
+Before=docker.service
 Documentation=https://github.com/openshift/origin
 
 [Service]
--- a/rel-eng/openshift-sdn-master.service
+++ b/rel-eng/openshift-sdn-master.service
@@ -2,7 +2,9 @@
 Description=OpenShift SDN Master
 Documentation=https://github.com/openshift/openshift-sdn
 After=openshift-master.service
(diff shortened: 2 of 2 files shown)


sdodson:
Based on the number of folks having trouble with docker starting up without lbr0 I've amended this to include ensuring openshift-sdn-node is started before docker.

sdodson:
@mrunalp @rajatchopra Do you mind reviewing this? We have a good number of people running into problems with beta2 and I'd like to get this in before beta3.

rel-eng/openshift-sdn-node.service
```diff
@@ -1,8 +1,13 @@
 [Unit]
 Description=OpenShift SDN Node
-Before=openshift-node.service
 After=openvswitch.service
+After=openshift-sdn-master.service
+After=openshift-master.service
+Wants=openshit-sdn-master.service
+Wants=openshift-master.service
```
rajatchopra: Why does the node-service want a master service?

sdodson: It's not necessary as it would be chained openshift-sdn-node wants openshift-sdn-master wants openshift-master. I was just trying to make each service describe it's entire chain. I'm happy to stick to immediate dependencies if you'd prefer.

rajatchopra: It confuses me at least, so just the immediate dependencies that solve the ordering problem would be nice.
@mrunalp your opinion?

rajatchopra: Actually on second thoughts, its harmless.

rajatchopra:
LGTM

mrunalp:
LGTM

PR #45, opened 2015-03-25
Normalize sysconfig between node and master

sdodson:
* Required values all have their own environment variable so that it's easier to use Ansible/Puppet etc against those values
* Optional values are to be appended to OPTIONS variable

### DIFF
--- a/rel-eng/openshift-sdn-master.service
+++ b/rel-eng/openshift-sdn-master.service
@@ -7,7 +7,7 @@ Requires=openshift-master.service
 [Service]
 Type=simple
 EnvironmentFile=-/etc/sysconfig/openshift-sdn-master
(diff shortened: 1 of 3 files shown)


knobunc:
The code seems reasonable, and the use-case seems plausible. My only concern is that it adds makes previously optional config values required. However, since it gives them all defaults, so that shouldn't be an upgrade concern. But one of the required config values is set to an empty string (for MINION_IP). Does that mean it must be set? Or is empty the value... if so, is it required?

sdodson:
I'd have to review this to figure out how relevant this is today. I'd prefer to just close it, what do you think?

knobunc:
That is fine with me.

sdodson:
This is believed to no longer be relevant.

PR #48, opened 2015-03-27
Service dependencies round2

sdodson:
Sorry for the second round of systemd service changes. @sosiouxme and I discussed this a good bit and we've arrived at this pattern of adding WantedBy to the [Install] section being a good way to allow optional dependencies.

### DIFF
--- a/rel-eng/openshift-sdn-node.service
+++ b/rel-eng/openshift-sdn-node.service
@@ -3,8 +3,6 @@ Description=OpenShift SDN Node
 After=openvswitch.service
 After=openshift-sdn-master.service
 After=openshift-master.service
-Wants=openshit-sdn-master.service
-Wants=openshift-master.service
 Requires=openvswitch.service
 Before=openshift-node.service
 Before=docker.service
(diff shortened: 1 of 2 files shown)


sosiouxme:
LGTM, any reason to delay?

rel-eng/openshift-sdn-node.service
```diff
@@ -3,8 +3,7 @@ Description=OpenShift SDN Node
 After=openvswitch.service
 After=openshift-sdn-master.service
 After=openshift-master.service
-Wants=openshit-sdn-master.service
-Wants=openshift-master.service
+After=etcd.service
```
rajatchopra: Why do we need etcd service ordered on node?

sdodson: It's my understanding that openshift-sdn-node still talks to etcd rather than the API. If etcd is installed and enabled on the same host as openshift-sdn-node we should start after it's been started.

sdodson:
@sosiouxme I think it's good to go.

PR #50, opened 2015-04-10
Add SyslogIdentifier=openshift-sdn-{master,node} respectively

sdodson:
This will allow rsyslog users to split master and node logs if they
wish to do so.


mrunalp:
LGTM

PR #51, opened 2015-04-14
WIP: Native docs

mrunalp:
@rajatchopra PTAL.

I have to add diagram and firewall rules next.

### DIFF
--- a/Documentation/native/README.md
+++ b/Documentation/native/README.md
@@ -1,0 +1,2 @@
+Native container routing using network infrastructure
+==========================================================
--- a/Documentation/native/README.md
+++ b/Documentation/native/README.md
@@ -23,3 +23,26 @@ Network Overview
 * Router has routes for each node, so it can direct to the right node.
 * Nodes don't need any changes when new nodes are added unless the network topology is modified.
 * IP forwarding is enabled on each node.
+
+
+Node setup
+-----------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Documentation/native/README.md
```diff
@@ -0,0 +1,64 @@
...
+* Enable IP forwarding on the router.
+```
+sysctl -w net.ipv4.ip_forward=1
+```
+* Add a route per node added to the cluster.
+```
+ip route add 11.11.1.0/24 via 192.168.2.1 dev p3p1 
+ip route add 11.11.2.0/24 via 192.168.3.3 dev p3p2 
+ip route add 11.11.3.0/24 via 192.168.3.4 dev p3p2 
+```
```
rajatchopra: Add a generic example rule to explain what the above three mean, maybe?
e.g. ip route add <node_subnet> via <node_ip_address> dev <interface through which node is L2 accessible>

Also, instead of the 'ip route add' maybe we should depict the routing table. e.g.
```
Destination     Gateway         Genmask          Use Iface
default         10.14.16.254    0.0.0.0                 em1
10.14.16.0      0.0.0.0         255.255.255.0      em1
```

mrunalp: Yeah, that ip route example line you have is really good. I added that. I think I prefer ip route rules as the route command is considered deprecated. If you feel strongly, I can add it :)

mrunalp:
Merging this for now.

Issue #52, opened 2015-04-20
should be using /etc/sysconfig/docker-network for network config


thoraxe:
There is a file called `/etc/sysconfig/docker-network` that should be used for the network-specific config components.

detiber:
instead of modifying /etc/sysconfig/docker we should be modifying /etc/sysconfig/docker-network instead.
The entries in /etc/sysconfig/docker-network are just appended to the docker options after the entries in /etc/sysconfig/docker and /etc/sysconfig/docker-storage.

- We will need to set `DOCKER_NETWORK_OPTIONS='-b=lbr0 --mtu=1450'` instead of the  OPTIONS var we are setting now (--selinux-enabled is not needed because it is set by default in /etc/sysconfig/docker)
- We will not be able to set the --insecure-registry flag here, since that is overriden with the value from /etc/sysconfig/docker, but we can adjust for that in the ansible playbooks.

eparis:
flannel does the following, first it calls a helper function which generates, at run time, the options you are hard coding into /etc/sysconfig/docker (https://github.com/openshift/openshift-sdn/blob/master/ovs-simple/bin/openshift-sdn-simple-setup-node.sh#L54)
```
ExecStartPost=/usr/libexec/flannel/mk-docker-opts.sh -k DOCKER_NETWORK_OPTIONS -d /run/flannel/docker
```
So that creates /run/flannel/docker with the env variable DOCKER_NETWORK_OPTIONS=   Note that our script is smart enough to prepend previous values in DOCKER_NETWORK_OPTIONS. So it sources /etc/sysconfig/docker-network, rather than overwrite it (https://github.com/coreos/flannel/blob/master/dist/mk-docker-opts.sh) ....


---

Next flannel has a config file:
```
# cat /usr/lib/systemd/system/docker.service.d/flannel.conf
[Service]
EnvironmentFile=-/run/flannel/docker
```

Which means that docker will source /run/flannel/docker.

rajatchopra:
Merged #54

PR #54, opened 2015-04-20
Set DOCKER_NETWORK_OPTIONS via /etc/sysconfig/docker-network

sdodson:


sdodson:
Fixes #52 I think testing now

rel-eng/openshift-sdn-node.sysconfig
```diff
@@ -55,4 +55,4 @@ OPTIONS=
 # The $DOCKER_OPTIONS variable is used to overwrite the $OPTIONS
 # variable in the /etc/sysconfig/docker file. This is a pretty ugly
 # hack, but whatevs
```
rajatchopra: @sdodson Minor one. Update the comment above?

rajatchopra:
Runs fine with manual testing. Will re-test with rpm.
Ready for merge post comment-fix above.

rel-eng/openshift-sdn-node.sysconfig
sdodson: Should be fixed now, also linked to more elegant solution for anyone willing to invest more time.

PR #55, opened 2015-04-20
StartNode: return if GetSubnets returns error

Miciah:
Change `controller.(*OvsController).StartNode` so that if it gets an error from `GetSubnets`, `StartNode` will return that error rather than continuing on to dereference a nil pointer and panic.

### DIFF
--- a/ovs-simple/controller/controller.go
+++ b/ovs-simple/controller/controller.go
@@ -200,6 +200,7 @@ func (oc *OvsController) StartNode(sync, skipsetup bool) error {
 		subnets, err := oc.subnetRegistry.GetSubnets()
 		if err != nil {
 			log.Errorf("Could not fetch existing subnets: %v", err)
+			return err
 		}
 		for _, s := range *subnets {
 			oc.AddOFRules(s.Minion, s.Sub)


sdodson:
Out of curiosity, under which conditions would you expect it to error? I'm currently hitting this in newer deployments using openshift v0.5

Miciah:
It may have been caused by my directly mucking with the registry to try to correct some internal/external address confusion.  I probably had put some invalid JSON in the registry.

pravisankar:
Related code was moved under ovssubnet/common.go and this issue was resolved.

knobunc:
Ravi: I saw that it moved, but I didn't see the proposed return in the changed version.  I'd like to add the proposed changed to common.go unless you object.

pravisankar:
Oops, GetSubnets() is called in 2 places. Yes, we need to return the error.
@knobunc please fix it.

knobunc:
Micah, will you reapply this to the new code and resubmit the patch?  I really want to get this in.  If you don't have time, let me know and I can take over.  Thanks!

Miciah:
@knobunc, yes, I can do that.  I'll get to it sometime this week.  If this is urgent, let me know, and I'll get to it sooner rather than later.

BTW, Micah and Miciah are distinct people; I am the latter person only.  *grin*.

knobunc:
Nope, not urgent... I was just working through the backlog.

pravisankar:
Will be fixed with this pr: https://github.com/openshift/openshift-sdn/pull/146

PR #60, opened 2015-04-27
first set of reorg for api-fication of openshift-sdn

rajatchopra:
Moved some types/structs from registry/ package to a new api/ package. This will facilitate making registry/ as one of the interface implementations. A pluggable implementation can then, choose a non-etcd based backend as an example.

@mrunalp Review please.


mrunalp:
LGTM

PR #61, opened 2015-04-28
remove revision arg from api as it is an etcd abstraction

rajatchopra:
@mrunalp Review please.
Minor change to omit revision number in API interface.


mrunalp:
LGTM

PR #63, opened 2015-05-04
specfile fixes

sdodson:
I was getting rpm build failures. Does openshift-ovs-subnet get installed into /bin/openshift-ovs-subnet or to /usr/libexec/kubernetes/kubelet-plugins/net/exec/redhat~openshift-ovs-subnet/openshift-ovs-subnet? I think the former?

### DIFF
--- a/openshift-sdn.spec
+++ b/openshift-sdn.spec
@@ -69,13 +69,14 @@ cp $(pwd)/ovssubnet/bin/openshift-ovs-subnet $(pwd)/openshift-ovs-subnet
 %install
 
 install -d %{buildroot}%{_bindir}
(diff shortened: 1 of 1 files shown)


rajatchopra:
Actually the former is really needed because that is what kubernetes picks up. The /bin/openshift-ovs-subnet is something we can skip actually.

sdodson:
Ok, I think I've fixed it then. The error was that during rpm build it was attempting to install into the root filesystem rather than relative to the buildroot.

rajatchopra:
Oh yea didnt prefix it with buildroot. Thanks. Looks good.

PR #65, opened 2015-05-08
Change minions path to /kubernetes.io/minions/

sdodson:
Storage paths were changed and also made to be configurable, /kubernetes.io/minions/ is the new default. This should probably be made configurable but given the other changes coming soon I've just done this for now.


sdodson:
Better fix in #66

PR #67, opened 2015-05-08
state file no more with docker 1.6, use ethtool instead. Credit: Dan Winship

rajatchopra:
@mrunalp @danwinship
Please review. Tested on the openshift cluster.


mrunalp:
LGTM

PR #69, opened 2015-05-14
Add ipvlan IPv4 support

dcbw:
ipvlan in general only works with 4.0 and later kernels.

ipvlan L2 + DHCP will only work reliably with 4.1 and later kernels,
and because ipvlan uses the same MAC address for all interfaces, the
DHCP server must support differentiating leases via Client ID.  One
notable exception is the VirtualBox DHCP server, which does not support
Client ID and thus will not work with ipvlan L2 + DHCP.

This patch changes the '-kube' option into a '-kubenet' option that
accepts a couple different networking plugins.  In addition to the
previous '-kubenet=kube' option this commit adds '-kubenet=ipvlan-l2'
and '-kubenet=ipvlan-l3' options.  In all ipvlan configurations
the pod gets a single ipvlan interface created as a child of the
physical NIC on the minion that has the defualt route.  The pod-private
IP address and a publicly routable IP address are both assigned to
the ipvlan interface in parallel with each other.

Pod-private IP addresses work the same way for all modes.  The only
difference is how the public IP address is assigned and what layer the
ipvlan operates in:

ipvlan-l2 + DHCP: the ipvlan operates at L2 and is aware of broadcast
and multicast frames from the physical LAN.  The public IP address is
retrieved from the physical LAN's normal DHCP server.  This requires
that the container image have the /sbin/ip and ISC dhclient binaries
available.

ipvlan-l2/ipvlan-l3 + static:  the public IP address is allocated from
a pool configured on the master using the --public-network option of
openshift-sdn like so:

`--public-network=<start>-<stop>/<prefix>+<gateway>`

The URL of the master's public network IPAM is saved in etcd and read
on from each minion when the pod is started.

### DIFF
--- a/ipvlan/bin/openshift-ipvlan-kube-subnet-setup.sh
+++ b/ipvlan/bin/openshift-ipvlan-kube-subnet-setup.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -ex
+
+## docker
+if [[ -z "${DOCKER_OPTIONS}" ]]
+then
+    DOCKER_OPTIONS='-b=none --mtu=1450 --selinux-enabled'
+fi
+
+if ! grep -q "^OPTIONS='${DOCKER_OPTIONS}'" /etc/sysconfig/docker
+then
+    cat <<EOF > /etc/sysconfig/docker
+# This file has been modified by openshift-sdn. Please modify the
+# DOCKER_OPTIONS variable in the /etc/sysconfig/openshift-sdn-node,
+# /etc/sysconfig/openshift-sdn-master or /etc/sysconfig/openshift-sdn
+# files (depending on your setup).
+
+OPTIONS='${DOCKER_OPTIONS}'
+EOF
+fi
+systemctl daemon-reload
+systemctl restart docker.service
+
+mkdir -p /etc/openshift-sdn
--- a/pkg/pubnet_ipam/common.go
+++ b/pkg/pubnet_ipam/common.go
@@ -1,0 +1,187 @@
+package ovs_subnet
+
+import (
+	"fmt"
+	log "github.com/golang/glog"
+	"net"
+	"sync"
+	"strings"
+	"path"
+
+	"github.com/openshift/openshift-sdn/pkg/netutils"
+	"github.com/coreos/go-etcd/etcd"
+
+	netutils_server "github.com/openshift/openshift-sdn/pkg/netutils/server"
+)
+
+type PubnetIpam struct {
+	pnIpam          *netutils.IPAllocator
+	gateway		string
+	mux             sync.Mutex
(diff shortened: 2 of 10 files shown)


rajatchopra:
We should probably rename the openshift-ipvlan.. scripts to openshift-sdn-ipvlan..
Just namespacing. Need to do that for openshift-ovs-subnet too.

ipvlan/bin/openshift-ipvlan-kube-subnet-setup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -ex
+
+## docker
+if [[ -z "${DOCKER_OPTIONS}" ]]
+then
+    DOCKER_OPTIONS='-b=none --mtu=1450 --selinux-enabled'
```
rajatchopra: Should the mtu be 1450 still?

pkg/pubnet_ipam/common.go
```diff
@@ -0,0 +1,187 @@
...
+	etcdCli		*etcd.Client
+	etcdPath	string
+}
+
+func NewPubnetIpamServer(publicNetwork string,
+                         etcdPeers []string,
+                         etcdPath string,
+                         etcdCertfile string,
+                         etcdKeyfile string,
+                         etcdCafile string) (*PubnetIpam, error) {
```
rajatchopra: The etcd portion needs to be split out into an api and a registry. The api will define the methods needed and the registry will actually implement using etcd.

Just the same way as ovssubnet uses the pkg/api/types.go:<SubnetRegistry>
Create a pkg/api/types.go:<IpvlanRegistry>

dcbw:
"We should probably rename the openshift-ipvlan.. scripts to openshift-sdn-ipvlan..
Just namespacing. Need to do that for openshift-ovs-subnet too."

I stayed away from "sdn" here for ipvlan because ipvlan isn't really SDN-related.  Can change if you like though?

Also, is "subnet" really appropriate for the ipvlan stuff here?

ipvlan/bin/openshift-ipvlan-kube-subnet-setup.sh
dcbw: This was copied from ovssubnet/bin/openshift-sdn-kube-subnet-setup.sh and ovssubnet/bin/openshift-sdn-simple-setup-node.sh.  Was the 1450 done to account for overlay network (vxlan) headers?  If so, then I'll remove the MTU bit here since ipvlan doesn't use overlay at all, plus it doesn't use the docker bridge either.

rajatchopra: Yes 50 bytes for the overlay header. Yes, meaningless if b=none.

PR #70, opened 2015-05-15
Add ipvlan IPv4 support

dcbw:
ipvlan in general only works with 4.0 and later kernels.

ipvlan L2 + DHCP will only work reliably with 4.1 and later kernels,
and because ipvlan uses the same MAC address for all interfaces, the
DHCP server must support differentiating leases via Client ID. One
notable exception is the VirtualBox DHCP server, which does not support
Client ID and thus will not work with ipvlan L2 + DHCP.

This patch changes the '-kube' option into a '-kubenet' option that
accepts a couple different networking plugins. In addition to the
previous '-kubenet=kube' option this commit adds '-kubenet=ipvlan-l2'
and '-kubenet=ipvlan-l3' options. In all ipvlan configurations
the pod gets a single ipvlan interface created as a child of the
physical NIC on the minion that has the defualt route. The pod-private
IP address and a publicly routable IP address are both assigned to
the ipvlan interface in parallel with each other.

Pod-private IP addresses work the same way for all modes. The only
difference is how the public IP address is assigned and what layer the
ipvlan operates in:

ipvlan-l2 + DHCP: the ipvlan operates at L2 and is aware of broadcast
and multicast frames from the physical LAN. The public IP address is
retrieved from the physical LAN's normal DHCP server. This requires
that the container image have the /sbin/ip and ISC dhclient binaries
available.

ipvlan-l2/ipvlan-l3 + static: the public IP address is allocated from
a pool configured on the master using the --public-network option of
openshift-sdn like so:

`--public-network=<start>-<stop>/<prefix>+<gateway>`

The URL of the master's public network IPAM is saved in etcd and read
on from each minion when the pod is started.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,12 @@ install:
 	mkdir -p /usr/libexec/kubernetes/kubelet-plugins/net/exec/redhat~openshift-ovs-subnet/
 	cp -f $(OUT_DIR)/local/go/bin/openshift-ovs-subnet /usr/libexec/kubernetes/kubelet-plugins/net/exec/redhat~openshift-ovs-subnet/
 
+	cp -f $(OUT_DIR)/local/go/bin/openshift-ipvlan-subnet /usr/bin/
+	cp -f $(OUT_DIR)/local/go/bin/openshift-ipvlan-kube-subnet-setup.sh /usr/bin/
+	mkdir -p /usr/libexec/kubernetes/kubelet-plugins/net/exec/redhat~openshift-ipvlan-subnet/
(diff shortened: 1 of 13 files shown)


dcbw:
Sorry about the new pull request, when working on topic branches I usually kill the upstream branch and repush (since I do a lot of rebasing for fixups) and most non-github servers deny a push when it's not a simple fast-forward of what's already there...

dcbw:
This request should fix all of Rajat's comments on the original branch plus fixes the Go formatting per hack/verify-gofmt.sh

dcbw:
This is going to need a major rework at some point in the future, but we don't have time to do that right now.  Closing this PR for now.

PR #72, opened 2015-05-19
Update /etc/sysconfig/docker-network rather than /etc/sysconfig/docker

sdodson:
Similar fix as to that in #54


rajatchopra:
LGTM wait for green from Travis.

PR #74, opened 2015-05-28
Add iptables rule to allow forwarding from clusternetwork

sdodson:
still testing this


sdodson:
WORKSFORME so far, though, what's up with all those any/any rules above it?

```
[root@ose3-node2 ~]# iptables -L FORWARD
Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         
DOCKER     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere             ctstate RELATED,ESTABLISHED
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             10.1.0.0/16         
REJECT     all  --  anywhere             anywhere             reject-with icmp-host-prohibited
```

sdodson:
@mrunalp can you take a look at this soon'ish?

mrunalp:
@sdodson It looks good. I will merge it once you confirm that it works in your testing. Thanks!

mrunalp:
nvm, looks like it is working for you. LGTM. @rajatchopra could you take a look as well?

sdodson:
Hmm, I'm still getting networking problems. Testing some more

sdodson:
FORWARD -s ${cluster_subnet} -j ACCEPT

Are we sure we have the rules right?

rajatchopra:
@sdodson How are we sure the reject one is the last line? Maybe we should grep for 'reject'?
We can merge this, but it maybe insufficient.

sdodson:
Updated to accept forwarding to/from cluster_network and updated the logic to find the reject rule.

PR #79, opened 2015-06-04
Set net.bridge.bridge-nf-call-iptables=0 during init

sdodson:


sdodson:
I wasn't sure exactly where in the order it should be, but given it works when you set it after the fact I imagine it's safe anywhere.

PR #82, opened 2015-06-05
Fix origin issue#2834; submit existing subnets as in-use

rajatchopra:
Do not rely on InitMinions implementation to know if subnets already exist or not.
InitMinions always returns nil error with OpenShift integration e.g.

@mrunalp Review please

Fixes: https://github.com/openshift/origin/issues/2834

### DIFF
--- a/ovssubnet/common.go
+++ b/ovssubnet/common.go
@@ -90,17 +90,15 @@ func (oc *OvsController) StartMaster(sync bool, containerNetwork string, contain
 	}
 
 	// initialize the subnet key?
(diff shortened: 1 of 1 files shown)


mrunalp:
@rajatchopra Were you able to test with a new cluster as well?

rajatchopra:
Vagrant up works. So I guess it is good to go. I will test some more in a while and we can revert/add-another-fix if I find something off.

mrunalp:
SGTM.

PR #83, opened 2015-06-09
Signal when the SDN is ready

smarterclayton:
Part of openshift/origin#2655

Already included as an UPSTREAM patch


smarterclayton:
Hrm, SeenAllSources is broken in this.

PR #85, opened 2015-06-10
sdn should not re-configure docker/ovs if it appears to be a harmless restart

rajatchopra:


rajatchopra:
@mrunalp This is ready to merge. Upstream merge has already happened.

PR #87, opened 2015-06-13
WIP: get rid of bash scripts

eparis:
I wrote and iptables, ovs-{vs,of}ctl, brctl, and ip control modules. I then ported most of openshift-sdn-kube-blah.sh into kube.go.  I can finish it up soon...

This does not work right now!  I'm not calling what is left of the bash, but i'm not dong everything the bash was doing...


eparis:
still haven't actually tried it, but this should (might) get rid of the bash script entirely...

danwinship:
#166 is a revision/reimplementation of this

PR #89, opened 2015-06-16
fix issue#88 - generate unique cookie based on vtep

rajatchopra:
Sample output of hex encoding the IP address:
Checksum '192.168.1.4': c0a80104
Checksum '192.168.1.5': c0a80105
Checksum '10.168.1.5': 0aa80105

As long as the IP address of a node is unique (which should always be the case), the cookie will be unique too.

@mrunalp Review please.


mrunalp:
LGTM

PR #92, opened 2015-06-18
fix error in bash script

rajatchopra:
@mrunalp review please


ovssubnet/bin/openshift-ovs-subnet
```diff
@@ -26,7 +26,7 @@ Setup() {
 
     pid=$(docker inspect --format "{{.State.Pid}}" ${net_container})
     network_mode=$(docker inspect --format "{{.HostConfig.NetworkMode}}" ${net_container})
-    if [ "$network_mode"=="host" ]; then
+    if [ "$network_mode" == "host" ]; then
```
mrunalp: Let's add {} as well.

PR #93, opened 2015-06-29
move api,types,bin files to respective controllers

rajatchopra:
@mrunalp PR to the multitenant branch.
Just moves the bin/ scripts to respective controllers. Also moves the api/types.go and registry/registry.go into ovssubnet/ because the api and registry is specific to that only.


mrunalp:
LGTM

PR #104, opened 2015-07-09
make cross node services work

rajatchopra:
Services need to work across nodes. That means being able to reach a pod (any namespace) from any host. The originating node's OVS rules do not handle a swift exit for return traffic (vnid != 0) meant for tun0 (port 2, IP: subnet_gateway). 

This PR adds a special rule in table 2 (as all input traffic from vxlan port is directed to table 2) to send anything meant for the host/service/tun0 directly out (no need to flood the rest of the ports).
Could have put this rule in table 5 also where all the rest of ports decide on which packet is whose, but why jump an extra table.

Also add priority to rules where the matching criterion can overlap (do not assume insertion order).

@dcbw : Review please. Services from another host didnt work without this and they work with this fix.


dcbw:
LGTM; I guess otherwise we'd have to depend on the learn action putting a rule in table 7 for tun0, which seems pointless since we know where traffic for 10.1.x.1 should go

PR #107, opened 2015-07-13
multitenant cleanups; no actual bug fixes

danwinship:
Also, not tested because I'm not sure how to test the services stuff? We need to automate our testing...


dcbw:
LGTM

PR #115, opened 2015-07-29
service isolation support

danwinship:
Implementation of the basis of service isolation; does not actually implement it; it will only be implemented in origin

### DIFF
--- a/ovssubnet/registry/registry.go
+++ b/ovssubnet/registry/registry.go
@@ -151,6 +151,14 @@ func (sub *EtcdSubnetRegistry) GetMinions() (*[]string, error) {
 	return &minions, nil
 }
 
+func (sub *EtcdSubnetRegistry) InitServices() error {
+	return nil
+}
+
+func (sub *EtcdSubnetRegistry) GetServices() (*[]api.Service, error) {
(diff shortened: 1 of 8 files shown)


rajatchopra:
LGTM
@dcbw Review please

ovssubnet/registry/registry.go
```diff
@@ -230,6 +238,11 @@ func (sub *EtcdSubnetRegistry) GetContainerNetwork() (string, error) {
 	return resp.Node.Value, err
 }
 
+func (sub *EtcdSubnetRegistry) GetServicesNetwork() (string, error) {
+	// FIXME
+	return "172.30.0.0/16", nil
```
pravisankar: You can use existing 'servicesSubnet' master config option for this one.
pkg/cmd/server/api/v1/types.go: ServicesSubnet string `json:"servicesSubnet"`

danwinship: That's only available on the master (where it can be overridden with a command-line argument). We need it on the minions.

rajatchopra: Do we need to create a resource for it?

PR #121, opened 2015-08-06
Fix veth lookup for kernels with interface suffixes

marun:
openshift-ovs-subnet fails on Fedora 22 due to 'ip link show' displaying
interface names with a suffix starting with '@', e.g.

  24: vetha06b051@if23: ...

Invoking brctl or ovs-* with the interface 'vetha06b051@if23' fails, and
it is necessary to use the name without the suffix (e.g. 'vetha06b051').

### DIFF
--- a/ovssubnet/controller/kube/bin/openshift-ovs-subnet
+++ b/ovssubnet/controller/kube/bin/openshift-ovs-subnet
@@ -15,6 +15,17 @@ lockwrap() {
     ) 200>${lock_file}
 }
 
+# Retrieve the name of the host-local member of the veth pair that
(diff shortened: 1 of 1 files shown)


rajatchopra:
Looks Good to me.
@danwinship Quick review from you please

dcbw:
For context, this is/was a kernel change along with iproute changes; it's not specific to Fedora 22.  Also note that the interface won't always be suffixed with the parent interface either, so the regex should take that into account (as I'm sure it already does...)

marun:
@danwinship I've fixed the regex and formatting as per your comments.

danwinship:
lgtm

Issue #116, opened 2015-07-29
net.ipv4.ip_forward unset


mrunalp:
Saw another issue where net.ipv4.ip_forward was 0. We should set it in the setup script if we already aren't.

mrunalp:
@pravisankar ping

pravisankar:
Fixed in https://github.com/openshift/openshift-sdn/pull/122

PR #122, opened 2015-08-07
Enable IP forwarding during openshift sdn setup

pravisankar:
Closes: #116


rajatchopra:
@dcbw could you review this please.
Specifically is it better that we set ipv4.conf.all.forwarding = 1? Or shall we do it specifically?

Issue #120, opened 2015-08-04
Calculate the pmtu instead of hardcoding it


rajatchopra:
Currently the mtu is hardcoded to 1450. While this works for most cases, its not valid when the assumed underlay is an overlay itself. Use tracepath to reach the master from the node and subtract 50 from it?

jeremyeder:
Performance team decided to shy away from PTMU in the network-throughput tuned profile, collectively felt that changing the MTU on an interface would really surprise sysadmins.  Our implementation idea was PMTU to the default gateway...where else would you measure to?

OpenStack does have something interesting though:  they auto-set the Neutron interface to DHCP-MTU-minus 50.

So, we might need 2 solutions; one for OSE and we might get it for free when running on OpenStack.  Or perhaps adopt OpenStack technique for OSE.  This could be an interesting topic for the f2f.

pravisankar:
We have decided not to change the MTU, default value will be 1450 but there will be config option for the admin to fine tune the value based on their need (like 9000 for jumbo frames).

jeremyeder:
Cool, yeah.   If a user needs the throughput improvement from jumbo frames, they probably want to get that data off of the overlay network.  If they're CPU bottlenecked in a packets-per-second type of workload, that's a separate issue.

pravisankar:
Fix merged.

PR #123, opened 2015-08-10
Make SDN MTU configurable

pravisankar:
closes: #120

### DIFF
--- a/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
+++ b/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
@@ -8,7 +8,8 @@ subnet=$2
 cluster_subnet=$3
 subnet_mask_len=$4
 tun_gateway=$5
-printf 'Container network is "%s"; local host has subnet "%s" and gateway "%s".\n' "${cluster_subnet}" "${subnet}" "${subnet_gateway}"
+mtu=$6
+printf 'Container network is "%s"; local host has subnet "%s", mtu "%d" and gateway "%s".\n' "${cluster_subnet}" "${subnet}" "${mtu}" "${subnet_gateway}"
 TUN=tun0
 
 # Synchronize code execution with a file lock.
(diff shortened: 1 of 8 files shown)


pravisankar:
@rajatchopra PTAL

ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
```diff
@@ -125,6 +126,12 @@ EOF
     sysctl -w net.ipv4.ip_forward=1
     sysctl -w net.ipv4.conf.${TUN}.forwarding=1
 
+    # tcp_mtu_probing controls TCP packetization-layer path MTU discovery
+    # Value: 1 - Disabled by default, enabled when an ICMP black hole detected
+    sysctl -w net.ipv4.tcp_mtu_probing=1
+    # Recommended value as per RFC#4821
+    sysctl -w net.ipv4.tcp_base_mss=1024
+
```
rajatchopra: Are these flags always needed?

pravisankar: These flags will help in case of black hole.

sdodson: Do we want to remove the option to configure this via the
DOCKER_NETWORK_SETTINGS envar? If that envar is set these settings won't
take effect and that could be confusing, if we do that then we'd want to
make sure that decide whether or not to include the bridge name as well.

On Mon, Aug 10, 2015 at 7:14 PM, Ravi Sankar Penta <notifications@github.com

ovssubnet/controller/kube/kube.go
```diff
@@ -22,11 +22,11 @@ func NewFlowController() *FlowController {
 	return &FlowController{}
 }
 
-func (c *FlowController) Setup(localSubnet, containerNetwork, servicesNetwork string) error {
+func (c *FlowController) Setup(localSubnet, containerNetwork, servicesNetwork string, mtu uint) error {
 	_, ipnet, err := net.ParseCIDR(localSubnet)
 	subnetMaskLength, _ := ipnet.Mask.Size()
 	gateway := netutils.GenerateDefaultGateway(ipnet).String()
-	out, err := exec.Command("openshift-sdn-kube-subnet-setup.sh", gateway, ipnet.String(), containerNetwork, strconv.Itoa(subnetMaskLength), gateway).CombinedOutput()
+	out, err := exec.Command("openshift-sdn-kube-subnet-setup.sh", gateway, ipnet.String(), containerNetwork, strconv.Itoa(subnetMaskLength), gateway, fmt.Sprint(mtu)).CombinedOutput()
```
danwinship: awkward that we're using both strconv.Itoa() and fmt.Sprint() to convert integers to strings in the same line...

PR #125, opened 2015-08-11
Add support for dind in multitenant plugin

marun:
This change updates the multitenant setup script to support
docker-in-docker deployment, which uses supervisord instead of systemd.


marun:
cc: @rajatchopra

marun:
Once this merges, can the dind-supporting changes be pulled into origin so that dind deployment can work by default without an openshift-sdn repo?

PR #127, opened 2015-08-13
Multi-tenant service isolation (again!)

danwinship:
OK, resubmitting the service isolation stuff. You can diff against danwinship/openshift-sdn service-isolation-old to see what's changed; basically it's just adding the service network to StartMaster and WriteNetworkConfig, which is needed in the origin osdn plugin to get that into the ClusterNetwork.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,7 +17,7 @@ import (
 )
 
 type NetworkManager interface {
-	StartMaster(sync bool, containerNetwork string, containerSubnetLength uint) error
+	StartMaster(sync bool, containerNetwork string, containerSubnetLength uint, serviceNetwork string) error
 	StartNode(sync, skipsetup bool) error
 	Stop()
 }
@@ -25,6 +25,7 @@ type NetworkManager interface {
 type CmdLineOpts struct {
 	containerNetwork      string
 	containerSubnetLength uint
+	serviceNetwork        string
 	etcdEndpoints         string
 	etcdPath              string
 	etcdKeyfile           string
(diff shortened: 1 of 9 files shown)


danwinship:
@rajatchopra @pravisankar PTAL

ovssubnet/api/types.go
```diff
@@ -65,3 +69,23 @@ type NamespaceEvent struct {
...
+const (
+	TCP ServiceProtocol = "TCP"
+	UDP ServiceProtocol = "UDP"
+)
+
+type Service struct {
+	Name      string
+	Namespace string
+	IP        string
+	Protocol  ServiceProtocol
```
rajatchopra: Do we need the Protocol today?

danwinship: Yes, AddServiceOFRules() adds either ",tcp" or ",udp" to the flow rule based on that

main.go
```diff
@@ -158,7 +158,7 @@ func main() {
 			log.Fatalf("Failed to start openshift sdn in node mode: %v", err)
 		}
 	} else if opts.master {
-		err := be.StartMaster(opts.sync, opts.containerNetwork, opts.containerSubnetLength)
+		err := be.StartMaster(opts.sync, opts.containerNetwork, opts.containerSubnetLength, "172.30.0.0/16")
```
pravisankar: Instead of hard coding to "172.30.0.0/16", can we make this as config option and use "172.30.0.0/16" as default value if not explicitly specified? (similar to containerNetwork)

ovssubnet/common.go
```diff
@@ -312,7 +314,12 @@ func (oc *OvsController) StartNode(sync, skipsetup bool) error {
 			log.Errorf("Failed to obtain ContainerNetwork: %v", err)
 			return err
 		}
-		err = oc.flowController.Setup(oc.localSubnet.Sub, containerNetwork)
+		servicesNetwork, err := oc.subnetRegistry.GetServicesNetwork()
```
pravisankar: We should only do this in case on multinent solution.
Refer https://github.com/danwinship/openshift-sdn/blob/service-isolation/ovssubnet/common.go#L334

ovssubnet/common.go
```diff
@@ -336,6 +343,15 @@ func (oc *OvsController) StartNode(sync, skipsetup bool) error {
...
 
+	services, err := oc.subnetRegistry.GetServices()
+	if err != nil {
+		log.Errorf("Could not fetch existing services: %v", err)
+	}
+	for _, svc := range *services {
+		oc.flowController.AddServiceOFRules(oc.VNIDMap[svc.Namespace], svc.IP, svc.Protocol, svc.Port)
+	}
+	go oc.watchServices()
+
```
pravisankar: Valid only for multi-tenant flow controller otherwise VNIDMap won't be populated.

pravisankar:
@rajatchopra Please check the OF rules

ovssubnet/common.go
danwinship: flowController.Setup() has a servicesNetwork arg now though, so we have to pass something, even in the non-multitenant case. And GetServicesNetwork() will return the right answer whether we're using multitenant or not. So even though kube and lbr don't use it, it doesn't hurt (and they could theoretically use it in the future).

pravisankar: Okay, that makes sense.

Issue #131, opened 2015-08-18
Perform SDN setup when switching between Flat and Multitenant plugins


pravisankar:
Currently in the flat/multi-tenant setup scripts, setup_required() is same in both the cases and this causes the user to explicitly delete the lbr0 bridge when switching between flat and multi-tenant plugins and vice versa.
Enhance setup_required() in both the plugins to detect setup requirement when user switches between the plugins.

rajatchopra:
This is a broad issue, and somewhat talked about here: https://github.com/openshift/origin/issues/3940
We need to split the setup file anyway to accommodate docker daemon restarting vs OVS rules.

pravisankar:
Fixed in https://github.com/openshift/openshift-sdn/pull/181

Issue #111, opened 2015-07-15
ensure that setup is idempotent; use case: openshift-sdn in a container


rajatchopra:
There is logic today that if the docker linux bridge is already configured correctly, then the setup phase on the node is skipped. Ensure that this 'skip' still completes the steps that could be left behind by a possible earlier setup exit because docker daemon was restarted.

The scenario exists in the case where openshift-sdn-node is run as a docker container (privileged, of course). As soon as the 'setup sdn' restarts docker, the sdn process will end up killing itself leaving possibly unfinished tasks of the setup.

marun:
Might a better title for this issue be 'ensure that setup is idempotent'?  i.e. It should be possible to run network setup multiple times and be assured of the correct outcome if it completes without error.

PR #132, opened 2015-08-18
Detect SDN setup requirement when user switches between flat and multitenant plugins

pravisankar:
closes: #131


pravisankar:
@danwinship @rajatchopra PTAL

rajatchopra:
This helps with migration but is a short term fix. Closing this out in favour of better documentation and fix for https://github.com/openshift/openshift-sdn/issues/111

PR #133, opened 2015-08-19
Get node IP from GetNodes() instead of net.Lookup()

pravisankar:
This will avoid cases where node IP resolution on
master and node differs for various reasons (stale dns cache, etc.)

### DIFF
--- a/ovssubnet/common.go
+++ b/ovssubnet/common.go
@@ -64,19 +64,10 @@ func NewDefaultController(sub api.SubnetRegistry, hostname string, selfIP string
 
 func NewController(sub api.SubnetRegistry, hostname string, selfIP string, ready chan struct{}) (*OvsController, error) {
 	if selfIP == "" {
-		addrs, err := net.LookupIP(hostname)
+		var err error
+		selfIP, err = getNodeIP(hostname)
 		if err != nil {
-			log.Errorf("Failed to lookup IP Address for %s", hostname)
 			return nil, err
-		}
-		for _, addr := range addrs {
-			if addr.String() != "127.0.0.1" {
-				selfIP = addr.String()
-				break
-			}
-		}
-		if selfIP == "" {
-			return nil, fmt.Errorf("failed to lookup valid IP Address for %s (%v)", hostname, addrs)
 		}
 	}
 	log.Infof("Self IP: %s.", selfIP)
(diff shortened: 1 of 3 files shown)


pravisankar:
@rajatchopra @dcbw @danwinship PTAL

ovssubnet/registry/registry.go
```diff
@@ -142,28 +132,35 @@ func (sub *EtcdSubnetRegistry) InitNodes() error {
 	return err
 }
 
-func (sub *EtcdSubnetRegistry) GetNodes() (*[]string, error) {
+func (sub *EtcdSubnetRegistry) GetNodes() (*[]string, *[]string, error) {
```
rajatchopra: Let's return node objects (create new if needed). Two arrays of strings seems odd.

ovssubnet/registry/registry.go
```diff
@@ -42,24 +42,14 @@ func newNodeEvent(action, key, value string) *api.NodeEvent {
...
-					}
-				}
-			}
+			nodeEvent.NodeIP = nodeIP
+			return nodeEvent
 		}
 	}
 
-	fmt.Printf("Error decoding node event: nil key (%s,%s,%s).\n", action, key, value)
+	fmt.Errorf("Error decoding node event: nil key (%s,%s,%s).\n", action, key, value)
```
danwinship: fmt.Errorf() returns an error object and doesn't print anything. I think you mean "fmt.Fprintf(os.Stderr, ...)", but we don't do that in most other places, so if we're going to start fixing that, it should probably be globally, in a separate PR

pravisankar: There is a lot of inconsistency in registry.go for reporting errors. I'll handle this in a separate PR.

pravisankar:
@rajatchopra @danwinship Updated, please review/merge

ovssubnet/common.go
```diff
@@ -446,3 +413,24 @@ func (oc *OvsController) Stop() {
 	close(oc.sig)
 	//oc.sig <- struct{}{}
 }
+
+func GetNodeIP(nodeName string) (string, error) {
```
rajatchopra: Should this be a public function? getNodeIP was good, right?

dcbw:
LGTM now

PR #136, opened 2015-08-19
Make Get/Watch interfaces for nodes/subnets/namespaces consistent

pravisankar:


pravisankar:
@rajatchopra @danwinship PTAL

marun:
Why is it necessary to return a pointer to a slice, given that a slice is a reference type (i.e. the cost to copy is negligible)?

pravisankar:
Agree, Copy cost is not worth optimizing here but I do prefer Get/Watch API methods on nodes/namespaces/subnets/net-namespaces to be consistent.

marun:
Is there a reason not to fix the other methods instead?

pravisankar:
Yes, that will be better..thanks. I'll update the PR.

pravisankar:
Updated.
GetNodes() interface is fixed as part of this PR: https://github.com/openshift/openshift-sdn/pull/133

danwinship:
LGTM

dcbw:
LGTM

PR #137, opened 2015-08-19
Consistent error logging in standalone registry/IP allocator

pravisankar:


pravisankar:
@rajatchopra @danwinship PTAL

ovssubnet/registry/registry.go
```diff
@@ -59,7 +58,7 @@ func newNodeEvent(action, key, value string) *api.NodeEvent {
 		}
 	}
 
-	fmt.Printf("Error decoding node event: nil key (%s,%s,%s).\n", action, key, value)
+	log.Errorf("Error decoding node event: nil key (%s,%s,%s).\n", action, key, value)
```
danwinship: The glog functions don't require a trailing "\n". We might as well be consistent about that too.

pkg/netutils/ip_allocator.go
```diff
@@ -20,11 +22,11 @@ func NewIPAllocator(network string, inUse []string) (*IPAllocator, error) {
 	for _, netStr := range inUse {
 		_, nIp, err := net.ParseCIDR(netStr)
 		if err != nil {
-			fmt.Println("Failed to parse network address: ", netStr)
+			log.Errorf("Failed to parse network address: ", netStr)
```
danwinship: need to either use Errorln or else a %s

pravisankar:
@danwinship Updated

Issue #129, opened 2015-08-17
[multitenant plugin] Handle existing and admin namespaces during startup


rajatchopra:
Today when we migrate from openshift-ovs-subnet (not multitenant) to the multitenant plugin, we ignore the existing namespaces of the running cluster. This means that all existing namespaces get the default vnid, i.e. 0. This renders all existing namespaces with admin connectivity powers (can read/be read from anyone).
The multitenant plugin should assign VNIDs to existing namespaces also.

There should, however, be a way to specify which namespaces be left as 'admin' namespaces.

pravisankar:
Fixed in https://github.com/openshift/openshift-sdn/pull/176
Existing namespaces are isolated when user migrates from flat to multitenant sdn and admin can turn on/off project isolation/global visibility within the cluster using 'oadm pod-network' CLI cmd.

PR #138, opened 2015-08-20
For multitenant plugin, handle both existing and admin namespaces

pravisankar:
closes: #129

### DIFF
--- a/ovssubnet/api/types.go
+++ b/ovssubnet/api/types.go
@@ -25,7 +25,9 @@ type SubnetRegistry interface {
 	GetSubnetLength() (uint64, error)
 	CheckEtcdIsAlive(seconds uint64) bool
 
+	GetNamespaces() ([]string, error)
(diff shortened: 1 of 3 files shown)


pravisankar:
@rajatchopra PTAL

pravisankar:
Yes, we should release allocated NetID when writing to etcd fails. This is what happens, when I do copy+paste :) 
Actually, I should make oc.AdminNamespaces instead of oc.adminNamespaces. Intention here is openshift master will have config param that tells which all namespaces should be treated as special/admin namespaces. If the config param is not explicitly set, it will default to ["default"].

pravisankar:
Removing WIP, tested these changes on multi-node vagrant setup.
@rajatchopra @danwinship updated, PTAL

rajatchopra:
Where is the default? Plan to set in origin/osdn code?

pravisankar:
Yes, default is set in origin sdn plugin: https://github.com/openshift/origin/pull/4370

PR #139, opened 2015-08-20
Use firewalld D-Bus API to configure firewall rules

### LINKED ISSUE #32: Look into replacing iptables rules by firewalld rules
Suggested by @Miciah

danwinship:
Closes #32

(Note: I didn't bother changing the lbr code.)

### DIFF
--- a/pkg/firewalld/firewalld.go
+++ b/pkg/firewalld/firewalld.go
@@ -1,0 +1,108 @@
+package firewalld
+
+import (
+	"fmt"
+
+	"github.com/godbus/dbus"
+)
+
+type Interface struct {
+	obj         *dbus.Object
+	reloadFuncs []func()
+}
+
+const (
+	IPv4 = "ipv4"
+	IPv6 = "ipv6"
+)
+
+const (
+	firewalldName      = "org.fedoraproject.FirewallD1"
+	firewalldPath      = "/org/fedoraproject/FirewallD1"
+	firewalldInterface = "org.fedoraproject.FirewallD1"
+)
+
+func New() *Interface {
+	bus, err := dbus.SystemBus()
+	if err != nil {
+		return &Interface{}
+	}
+
+	fw := &Interface{
+		obj: bus.Object(firewalldName, dbus.ObjectPath(firewalldPath)),
+	}
+
+	go fw.dbusSignalHandler(bus)
+
+	return fw
+}
+
+func (fw *Interface) IsRunning() bool {
+	if fw.obj == nil {
+		return false
+	}
+
+	var zone string
+	err := fw.obj.Call(firewalldInterface+".getDefaultZone", 0).Store(&zone)
+	return err == nil
+}
+
+func (fw *Interface) AddRule(ipv, table, chain string, priority int, rule []string) error {
+	return fw.obj.Call(firewalldInterface+".direct.addRule", 0, ipv, table, chain, int32(priority), rule).Store()
+}
+
+func (fw *Interface) EnsureRule(ipv, table, chain string, priority int, rule []string) error {
+	var exists bool
+
+	err := fw.obj.Call(firewalldInterface+".direct.queryRule", 0, ipv, table, chain, int32(priority), rule).Store(&exists)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return nil
+	}
+
+	return fw.AddRule(ipv, table, chain, priority, rule)
+}
+
+func (fw *Interface) AddReloadFunc(reloadFunc func()) {
+	fw.reloadFuncs = append(fw.reloadFuncs, reloadFunc)
+}
+
+func (fw *Interface) dbusSignalHandler(bus *dbus.Conn) {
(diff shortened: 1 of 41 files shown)


danwinship:
@rajatchopra @pravisankar PTAL. The travis-ci failure seems spurious. (It's claiming the "glog" package doesn't exist?)

ovssubnet/controller/kube/kube.go
```diff
@@ -72,6 +91,44 @@ func (c *FlowController) manageLocalIpam(ipnet *net.IPNet) error {
...
+		if err != nil {
+			return err
+		}
+		err = fw.EnsureRule(firewalld.IPv4, "filter", "INPUT", 0,
+			[]string{"-p", "udp", "-m", "multiport", "--dports", "4789", "-m", "comment", "--comment", "001 vxlan incoming", "-j", "ACCEPT"})
+		if err != nil {
+			return err
+		}
+		err = fw.EnsureRule(firewalld.IPv4, "filter", "INPUT", 0,
+			[]string{"-i", "tun0", "-m", "comment", "--comment", "traffic from docker for internet", "-j", "ACCEPT"})
```
pravisankar: Assumption here is TUN=tun0 in the setup.sh, can we pass tun0 as arg to openshift-sdn-kube-subnet-setup.sh and setupIptables() that will ensure we use the same value in both places.

ovssubnet/controller/kube/kube.go
```diff
@@ -72,6 +91,44 @@ func (c *FlowController) manageLocalIpam(ipnet *net.IPNet) error {
 	return nil
 }
 
+func (c *FlowController) SetupIptables(fw *firewalld.Interface, containerNetwork string) error {
+	if fw.IsRunning() {
```
marun: Consider making this block more data-driven:

	for _, x := range []struct {
		ipv   string
		table string
		chain string
		value int
		args  []string
	}{
		{firewalld.IPv4, "nat", "POSTROUTING", 0, []string{"-s", ...}}
	} {
		err := fw.EnsureRule(x.ipv, x.table, x.chain, x.value, x.args)
		if err != nil {
			return err
		}
	}

marun:
Is there a reason every single change is being duplicated across the kube and multitenant plugins?  Why not factor out the common stuff to simplify ongoing maintenance?

danwinship:
I copied it to kube just because it was trivial to do so in this case (basically just diff | patch). I assume that in general we're not extending the kube plugin with any new features. Maybe I should just drop that part of this patch.

ovssubnet/controller/kube/kube.go
danwinship: @pravisankar, we assume the ovs bridge is "br0" in both setup.sh and AddOFRules() as well... should we also pass that as an argument?

The arguments to the setup script are getting pretty confusing...

pravisankar: Yes, args to the setup script are exploding (i'm adding one for mtu). Since this is not critical, may be just add a TODO comment for now and we can deal with that when we replace ovs rules with libovsdb api.

rajatchopra:
@danwinship This is dependent on the upstream PR, isn't it? Firewalld will be disabled in kubernetes without that, right?

danwinship:
@rajatchopra hm... well, I guess until the upstream PR (https://github.com/kubernetes/kubernetes/pull/12396) is committed, kubernetes will require that firewalld is disabled, and so the firewalld-specific code in this PR would never get run. But right now openshift-sdn always adds iptables rules to work around firewalld even when firewalld isn't running (aka always), so this PR is an improvement over that in a way.

danwinship:
updated to address @marun's "data-driven" comment, and to remove firewalld support from the kube code; if we merge kube+multitenant more later we can make the firewalld support generic then

danwinship:
@rajatchopra poke

pkg/firewalld/firewalld.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return fw.AddRule(ipv, table, chain, priority, rule)
+}
+
+func (fw *Interface) AddReloadFunc(reloadFunc func()) {
+	fw.reloadFuncs = append(fw.reloadFuncs, reloadFunc)
+}
+
+func (fw *Interface) dbusSignalHandler(bus *dbus.Conn) {
```
rajatchopra: Can we re-use some of this code when kubernetes upstream PR goes in?

danwinship: That was my original plan, until I remembered that openshift-sdn doesn't depend on/include kubernetes...

If that changes, then yes, we could use kutil.iptables.AddReloadFunc() instead of this

rajatchopra:
Do we need to enhance the rpm spec? Any new dependencies that this will require? That list of files in Godeps is scary.

danwinship:
godbus is a purely-golang implementation of D-Bus, so no, there are no new dependencies

rajatchopra:
LGTM

Issue #134, opened 2015-08-19
TestIPServe fails sporadically


danwinship:
TestIPServe fails sporadically in Travis-CI:

    === RUN TestIPServe-2
    --- FAIL: TestIPServe-2 (0.00s)
    	server_test.go:32: Error in connecting to IPAM server
    FAIL
    exit status 1
    FAIL	github.com/openshift/openshift-sdn/pkg/netutils/server	0.005s

eg, https://travis-ci.org/openshift/openshift-sdn/builds/76176233, https://travis-ci.org/openshift/openshift-sdn/builds/76204528, https://travis-ci.org/openshift/openshift-sdn/builds/76209682

It's always the second run that fails... maybe something isn't getting cleaned up correctly after the first one?

pravisankar:
We didn't notice any TestIPServe failures recently, hoping https://github.com/openshift/openshift-sdn/issues/140 fixed it.
Closing the issue for now. We can reopen if it happens again.

danwinship:
Just failed again: https://travis-ci.org/openshift/openshift-sdn/jobs/79471475:

    === RUN TestIPServe-2
    --- FAIL: TestIPServe-2 (0.00s)
    	server_test.go:32: Error in connecting to IPAM server: %v Get http://127.0.0.1:9080/netutils/ip: dial tcp 127.0.0.1:9080: connection refused

(I already fixed the "%v" in git.)

PR #140, opened 2015-08-24
Improve error messages in pkg/netutils/server/server_test

danwinship:
Hopefully shedding some light on #134


danwinship:
OK, we apparently have some other new error, which is that Travis isn't automatically finding our dependencies any more (also seen in other failures today). The second patch fixes that.

PR #141, opened 2015-08-26
Revoke VNID for admin namespaces

pravisankar:


pravisankar:
@rajatchopra PTAL

danwinship:
makes sense to me

PR #144, opened 2015-09-01
Implement kubernetes status hook in kube plugin

### LINKED ISSUE #142: implement status hook in network plugins
origin's multitenant plugin currently returns nil for the Status() call, with a "TODO unbreak this" comment

When using the kube plugin, its exec script gets run, sees that the command is not "init", "setup", or "teardown", and exits with status 1. kubelet then logs an error:

    Aug 26 16:30:03 ose3-node1-ansible.test.example.com atomic-openshift-node[4554]: E0826 16:30:03.699146    4554 manager.go:313] NetworkPlugin redhat/openshift-ovs-subnet failed on the status hook for pod 'docker-registry-1-deploy' - exit status 1

although the end result (other than the logged error) is the same as in the multitenant case.

danwinship:
Pretty simple. The only trick is that we have to remove the global "set -x" since the exec plugin tries to parse the combined stderr+stdout for the status, so we can't use -x on the status hook.

Closes #142. (The multitenant plugin side is in origin, not openshift-sdn, and will be addressed by https://github.com/openshift/origin/issues/4170.)

### DIFF
--- a/ovssubnet/controller/kube/bin/openshift-ovs-subnet
+++ b/ovssubnet/controller/kube/bin/openshift-ovs-subnet
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -ex
+set -e
 
 lock_file=/var/lock/openshift-sdn.lock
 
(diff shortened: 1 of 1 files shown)


rajatchopra:
Looks good. And now we need a godeps rebase to origin. @pravisankar Is there one due already?

Issue #145, opened 2015-09-02
Handle race conditions during master/node sdn setup


pravisankar:
Current model during master/node sdn setup:
(1) Fetch existing stuff from etcd store (nodes, subnets, namespaces, network namespaces)
(2) Perform some stuff: bookkeeping like VNID map or adding openflow rules, etc.
(3) Watch for any new changes in etcd store

Between steps (1) and (3), if there are any changes to the etcd store,
that information is lost and we may end up with missing/stale OF rules, etc.
We need to synchronize Get/Watch methods for the objects to ensure
no events are missed, meaning if Get() returned some items till revision 'r',
we need to Watch() for any items after revision 'r'.

pravisankar:
Fixed in #146 and #4219

PR #146, opened 2015-09-03
Fix SDN race conditions during master/node setup

pravisankar:
Related Issue: #145 

Watch interface for SDN resources takes ready and startVersion channels.
ready channel provides synchronization between watch and get resource
startVersion channel tells watch interface from which resource version to process the changes. Without this channel, watch interface may process the items multiple times and that can have side effects. Example: delete namespace needs deletion of VNID but ignoring if not present in this case may miss genuine error in our code. 

Ignore reviewing  ovssubnet/registry/registry.go (stale code, changes made just to make 'make' happy)

### DIFF
--- a/ovssubnet/common.go
+++ b/ovssubnet/common.go
@@ -3,9 +3,11 @@ package ovssubnet
 import (
 	"errors"
 	"fmt"
-	log "github.com/golang/glog"
 	"net"
+	"strings"
 	"time"
+
+	log "github.com/golang/glog"
 
 	"github.com/openshift/openshift-sdn/ovssubnet/api"
 	"github.com/openshift/openshift-sdn/ovssubnet/controller/kube"
@@ -105,7 +107,7 @@ func (oc *OvsController) StartMaster(sync bool, containerNetwork string, contain
 	// initialize the subnet key?
 	oc.subnetRegistry.InitSubnets()
 	subrange := make([]string, 0)
-	subnets, err := oc.subnetRegistry.GetSubnets()
+	subnets, _, err := oc.subnetRegistry.GetSubnets()
 	if err != nil {
 		log.Errorf("Error in initializing/fetching subnets: %v", err)
 		return err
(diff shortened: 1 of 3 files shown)


danwinship:
You could get rid of the ready channels if you made the Watch functions be called synchronously and have them spin off a goroutine themselves after the setup part.

ovssubnet/common.go
```diff
@@ -123,13 +124,36 @@ func (oc *OvsController) StartMaster(sync bool, containerNetwork string, contain
...
+	waitForWatchReadiness(ready, "Node")
+
+	version, err := oc.serveExistingNodes()
 	if err != nil {
-		log.Warningf("Error initializing existing nodes: %v", err)
-		// no worry, we can still keep watching it.
+		return err
 	}
+	start <- version
+
```
danwinship: This pattern occurs 5 times. I think it should be possible to turn it into a function? (Maybe it would be awkward because of types.)

pravisankar:
@danwinship 
I thought about getting rid of ready channels for watch interface but decided to keep it for two reasons:
- If we need to pass args to create/init queue, it will not be possible to make watch ready beforehand.
  for example, passing namespace arg to watch services.
- Since watch has to be async operation, anyone writing their own subnet registry will not run into any issues. This watch interface is clean and don't need to expose local watch queues under OsdnRegistryInterface.

As suggested, I'll try to make a function for repeated patterns.

pravisankar:
@rajatchopra @danwinship Updated, PTAL

danwinship:
Looks good... though the abstraction didn't end up abstracting as much as I'd hoped. Feel free to use either the original code or the watchAndGetResource() version, whichever you like better.

(Although I still think it would be better with the "go"s moved into the watchFoo() functions, and then the watchFoo() functions returning the result of subnetRegistry.GetFoo() themselves after waitForWatchReadiness().)

PR #149, opened 2015-09-08
Status hook for multitenant plugin

pravisankar:


pravisankar:
@danwinship PTAL
Needed for https://github.com/openshift/origin/issues/4170

danwinship:
lgtm

PR #150, opened 2015-09-09
Fix race condition in pkg/netutils/server/server_test.go

### LINKED ISSUE #134: TestIPServe fails sporadically

danwinship:
TestIPServe() was running ListenAndServeNetutilServer() in a goroutine but not necessarily waiting for it to actually start listening. Fix that by having ListenAndServeNetutilServer() first listen and then
spawn a goroutine to serve, and then return, and having TestIPServe() wait for it to return.

Closes #134

(Not that we're actually using this code anyway any more... if we don't expect to use it in the future maybe we should just remove pkg/netutil/server ?)

### DIFF
--- a/ovssubnet/controller/kube/kube.go
+++ b/ovssubnet/controller/kube/kube.go
@@ -66,7 +66,6 @@ func (c *FlowController) manageLocalIpam(ipnet *net.IPNet) error {
 		return err
 	}
 	f.Close()
-	// listen and serve does not return the control
 	netutils_server.ListenAndServeNetutilServer(ipam, net.ParseIP(ipamHost), ipamPort, nil)
 	return nil
 }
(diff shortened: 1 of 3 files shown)


rajatchopra:
Good catch.
We will need the netutil/server when we have our own IPAM introduced. If we end up using another IPAM (something from flannel), then we can drop it.

PR #152, opened 2015-09-14
Cleanup docker0 interface before restarting docker

EricMountain-1A:
Docker creates docker0 with a 172.16/16 route by default, and does not clean it up on exit.  This interface is not used with OpenShift-SDN and the routing simply conflicts with existing hosts folks may have in this private IP range.

OpenShift-SDN creates a bridge (lbr0) for Docker to use, however it is brought up after the 1st start of the docker service on boot, hence docker0 is always created.  The sequence is as follows (for OpenShift):
- docker service (which ​creates docker0 b.c. lbr0 does not exist yet)
- openshift-node service (with openshift-sdn-ovs linked in)
    - starts /usr/bin/openshift-sdn-kube-subnet-setup.sh
        - (*) creates /run/openshift-sdn/docker-network with "-b lbr0 ...", and of course lbr0 amongst others
        - restarts docker

So it looks like the cleanest place to get rid of docker0 is from the (*) in the above sequence onwards, and is implemented by this patch.

### DIFF
--- a/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
+++ b/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
@@ -122,6 +122,10 @@ EOF
         sysctl -w net.bridge.bridge-nf-call-iptables=0
     fi
 
+    # Cleanup docker0 since docker won't do it
+    ip link set docker0 down || true
+    brctl delbr docker0 || true
+
     # enable IP forwarding for ipv4 packets
     sysctl -w net.ipv4.ip_forward=1
     sysctl -w net.ipv4.conf.${TUN}.forwarding=1
--- a/ovssubnet/controller/lbr/bin/openshift-sdn-simple-setup-node.sh
+++ b/ovssubnet/controller/lbr/bin/openshift-sdn-simple-setup-node.sh
@@ -67,3 +67,7 @@ EOF
 
 systemctl daemon-reload
 systemctl restart docker.service
+
+# Cleanup docker0 since docker won't do it
+ip link set docker0 down || true
+brctl delbr docker0 || true
--- a/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
+++ b/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
@@ -148,6 +148,10 @@ EOF
         sysctl -w net.bridge.bridge-nf-call-iptables=0
     fi
 
+    # Cleanup docker0 since docker won't do it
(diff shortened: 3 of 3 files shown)


danwinship:
So why do we create lbr0 rather than just using docker0? We don't want to change the IP on docker0? (Or docker won't notice if we change the IP on it after startup?)

It seems like it would be better to just get this right initially rather than doing it wrong and then cleaning up afterward.

EricMountain-1A:
Docker won't notice if we change the IP on docker0 after startup.  In any case that requires destroying the interface and recreating it AFAICT, and getting docker to take notice requires restarting it.

Now I guess we could do that and reuse docker0 to our ends - that would in some way eliminate the cleanup.  OTOH, setting up a whole new interface makes it clearer that we are not relying on Docker's own (which makes it clearer in people's mind, and can help differentiate situations where a bug might be due to docker itself or due to another component doing the network setup).

As to getting this right initially: there's a bit of a chicken-and-egg situation.  Currently, openshift-node requires docker to be running in order to start (o/w it quits not finding /var/run/docker.sock), and in order to take "-b lbr0", Docker needs that interface to be created beforehand.  We'd need to split up the interface creation and ensure it is done before Docker starts the 1st time - this requires determining the subnet_gateway, subnet_mask etc. which are supplied as input to openshift-sdn-kube-subnet-setup.sh by openshift-node, which it has to derive from the etcd master.  So we'd need to break off a chunk of openshift-node (sdn master sync part?) to get that to run beforehand and setup the bridge. Passing that bridge into docker would avoid it creating docker0 in the 1st place.  May well be the way to go, but doesn't seem totally trivial. ;-)

EricMountain-1A:
Amended the commit to delete docker0 after docker has been restarted as suggested by @lhuard1A.

pravisankar:
If we are deleting docker0 bridge, we should also do it for multitenant plugin (https://github.com/EricMountain-1A/openshift-sdn/blob/destroy-docker0/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh)

EricMountain-1A:
@pravisankar updated with your comment.  I also did it for the lbr plugin (https://github.com/EricMountain-1A/openshift-sdn/blob/destroy-docker0/ovssubnet/controller/lbr/bin/openshift-sdn-simple-setup-node.sh) for consistency.

rajatchopra:
LGTM. Ravi merge if you think this is ok.
@danwinship Good point, but we may just want to get rid of the linux bridge configuration altogether now that we have Status hook implemented. We can have our own IPAM. So whenever we are ready to do that, or this should ship.

pravisankar:
LGTM

PR #155, opened 2015-09-21
Make "default" an admin namespace

danwinship:
Not sure at what point this got broken since I'm pretty sure it used to work, but anyway, we're currently not actually using VNID 0 for namespace "default", so it doesn't work...

### DIFF
--- a/ovssubnet/common.go
+++ b/ovssubnet/common.go
@@ -85,7 +85,7 @@ func NewController(sub api.SubnetRegistry, hostname string, selfIP string, ready
 		VNIDMap:         make(map[string]uint),
 		sig:             make(chan struct{}),
 		ready:           ready,
-		AdminNamespaces: make([]string, 0),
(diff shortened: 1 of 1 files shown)


danwinship:
@rajatchopra @pravisankar

pravisankar:
We decided not to merge https://github.com/openshift/origin/pull/4370 and that caused this issue.
It will be good, if we let the plugins to decide the sane defaults for AdminNamespaces.
Instead of assigning 'default' namespace in ovssubnet/common.go, do it in multitenant.go (https://github.com/openshift/origin/pull/4370/files#diff-d8a8179e40bbd42ad8c40b0fb54e2b4eR36)
kc.AdminNamespaces = append(kc.AdminNamespaces, "default") before calling kc.StartMaster(...)

danwinship:
ok, so this would be an origin change then

Issue #130, opened 2015-08-17
Either support openshift-sdn standalone mode or get rid of unsupported code/feature


pravisankar:
openshift-sdn binary is currently not tested in the standalone mode and the lbr controller is not useful when run as part of openshift binary. Based on team/PM inputs, If we plan to support/maintain/test openshift-sdn in standalone mode, we should fix it otherwise get rid of the code related to unsupported feature.

PR #156, opened 2015-09-22
Drop "lbr" mode

danwinship:
Related to #130, this drops support for lbr / running without kubernetes.

Can --skip-setup and --sync go away as well? origin always passes "false" for both

@pravisankar


rajatchopra:
--skip-setup might be needed in the future but we can always re-introduce it.
--sync should go away as it is an insecure way of operating in a cluster.

danwinship:
Re-pushed with --sync and --skip-setup deprecated. I didn't remove them from StartMaster()/StartNode() to avoid needing to sync an API update with origin.
The travis-ci failure seems to already exist in master... I'll look into that

danwinship:
er, sorry, repushed with gofmt'ing. (I got confused because running "./hack/test.sh" showed errors, but those are errors that are supposed to be printed...)

dcbw:
LGTM though maybe just fix StartNode since we're always passing false,false ?  or was the intent to panic() in one place only?

pravisankar:
can we close this pr?
https://github.com/openshift/openshift-sdn/pull/159 handles lbr/standalone binary cleanup

danwinship:
obsoleted by #159

PR #157, opened 2015-09-22
Move vovsbr to br0 port 3 rather than 9

danwinship:
It's dumb that we use port 9 for vovsbr, and then containers get ports 3-8 and 10+...

Changing this in kube would require an upgrade path, but we don't promise upgradeability with multitenant yet.

### DIFF
--- a/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
+++ b/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
@@ -52,7 +52,7 @@ function setup() {
     ip link set vovsbr txqueuelen 0
 
     ovs-vsctl del-port br0 vovsbr || true
-    ovs-vsctl add-port br0 vovsbr -- set Interface vovsbr ofport_request=9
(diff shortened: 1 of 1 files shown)


pravisankar:
Also update kube setup and openshift-docs [...vovsbr is added to br0 (port 9)... in https://github.com/openshift/openshift-docs/blob/master/architecture/additional_concepts/sdn.adoc]

danwinship:
> Changing this in kube would require an upgrade path, but we don't promise upgradeability with multitenant yet.

After some discussion on IRC, Ravi and I realized this doesn't actually break compatibility, it just means that you could end up with a cluster where some nodes have vovsbr on port 3 and others have it on port 9. (But no one cares what port any other node has it on, so that's OK.) So we can update kube too.

pravisankar:
LGTM

danwinship:
(The comment about updating kube too was wrong, because after an update, if it got restarted without re-running setup.sh, then ovs would still be using port 9, but kube.go would think it was using port 3. So it got merged without that part.)

PR #160, opened 2015-09-23
Update setup_required to do the correct check for docker-in-docker

danwinship:
The docker-in-docker supervisord setup uses different config files than the systemd setup, but we were deciding whether setup was needed based on whether the systemd files were correct.

@marun


dcbw:
LGTM though do systemd managed and supervisord-managed *really*  use different quoting in the docker options file?  stupid if so but there's nothing we can do about it.

PR #161, opened 2015-09-24
debug.sh improvements

danwinship:
We weren't outputting any service-related data. This adds "oc get service" output, and also does pod-to-service connectivity checks (which required streamlining the pod-to-pod connectivity code to avoid having way too much cut+paste).


dcbw:
LGTM

PR #162, opened 2015-09-25
code cleanup (no substantive changes)

danwinship:
I always get confused about which variables have plain IP addresses and which ones have IP/mask strings...

### DIFF
--- a/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
+++ b/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
@@ -3,13 +3,13 @@
 set -ex
 
 lock_file=/var/lock/openshift-sdn.lock
-subnet_gateway=$1
-subnet=$2
-subnet_mask_len=$3
-cluster_subnet=$4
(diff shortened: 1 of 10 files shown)


pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
```diff
@@ -3,13 +3,13 @@
...
-subnet_gateway=$1
-subnet=$2
-cluster_subnet=$3
-subnet_mask_len=$4
-tun_gateway=$5
+local_subnet_gateway=$1
+local_subnet_cidr=$2
+local_subnet_mask_len=$3
+cluster_network_cidr=$4
+service_network_cidr=$5
```
pravisankar: service_network_cidr not used? is it passed to this script to make it consistent with multitenant setup?

danwinship: Yes (the commit message mentions that, although github tends to make you not notice commit messages in reviews...)

pravisankar:
LGTM

PR #170, opened 2015-09-30
omit reconstructing IP address with docker inspect

rajatchopra:
Outputting nothing will lead to the default from 'docker inspect' itself. So avoid multiple pounding of docker daemon with docker inspect calls.

### DIFF
--- a/pkg/ovssubnet/controller/kube/bin/openshift-ovs-subnet
+++ b/pkg/ovssubnet/controller/kube/bin/openshift-ovs-subnet
@@ -67,12 +67,7 @@ Teardown() {
 }
 
 Status() {
-    ipaddr=$(docker inspect --format "{{.NetworkSettings.IPAddress}}" ${net_container})
-    if [ -z "$ipaddr" ]; then
-	exit 1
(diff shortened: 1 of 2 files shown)


rajatchopra:
@danwinship Review please

danwinship:
OK, so it looks like kubernetes/pkg/kubelet/dockertools/manager.go first calls dm.client.InspectContainer() (which does a "docker inspect" via the REST API), then calls the Status() hook, and if the status hook doesn't return anything, it pulls the NetworkSettings.IPAddress out of the InspectContainer() results. So doing the docker inspect ourselves in the plugins is completely pointless at this point, and getting rid of it is fine.

But this still leaves the problem that kubernetes is still hammering docker with InspectContainer() requests anyway... this isn't openshift-sdn's problem, but someone needs to figure out where that's coming from and cut it down to a more reasonable level.

jeremyeder:
Thanks guys, this will certainly help.  @danwinship wrt the kube hammer...yes.  Definitely, and it seems work on that is already underway.

PR #171, opened 2015-10-01
Simplify service tracking by using kapi.NamespaceAll

### LINKED ISSUE #158: nodes don't watch for services in pre-existing namespaces
@pravisankar 

Nothing ever calls watchServicesForNamespace() for namespaces that already exist when the node starts, so we never find out about any new services in those namespaces. In particular, nodes never find out about (non-pre-existing) services in "default".

Fixing this is somewhat complicated due to the jumping back and forth between origin/plugins/osdn.go and openshift-sdn/ovssubnet/common.go, and might be more easily fixed post-merge.

The ready/start stuff seems like it might be especially complicated in this case, since you have both the namespaces and the services within them to consider. Perhaps it would simplify things if we removed the Get() functions from the public APIs, and just had the Watch() functions, which would return (via the passed-in event channel) all pre-existing elements first, and then start returning new elements. Then the "start" info doesn't have to be passed all the way up and down the chain, and we don't have to do as much synchronization...

danwinship:
We don't need to track services per-namespace; we can just use the namespace kapi.NamespaceAll (aka ""), like "oc get --all-namespaces" does.

Closes #158

@pravisankar

### DIFF
--- a/plugins/osdn/osdn.go
+++ b/plugins/osdn/osdn.go
@@ -73,7 +73,7 @@ func (oi *OsdnRegistryInterface) CreateSubnet(nodeName string, sub *osdnapi.Subn
 }
 
 func (oi *OsdnRegistryInterface) WatchSubnets(receiver chan<- *osdnapi.SubnetEvent, ready chan<- bool, start <-chan string, stop <-chan bool) error {
-	eventQueue, startVersion := oi.createAndRunEventQueue("HostSubnet", nil, ready, start)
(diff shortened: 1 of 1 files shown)


rajatchopra:
Good stroke. I was searching for this one in the beginning but couldn't think of this obvious one.

Lgtm.

pravisankar:
Nice, clean solution.
lgtm

PR #172, opened 2015-10-02
More debug.sh updates

danwinship:
1. handle naming differences between origin and ose (https://bugzilla.redhat.com/show_bug.cgi?id=1265979)
2. Fix deprecation warnings since the kubernetes rebase
3. Improve an error message


dcbw:
LGTM

PR #173, opened 2015-10-02
Use iptables -w flag

danwinship:
To avoid problems with iptables locking, we need to use the -w flag, if available. Kubernetes's pkg/util/iptables has code to check for its availability and use it if so, so the simplest fix is to just switch to using that.

(Closes https://bugzilla.redhat.com/show_bug.cgi?id=1267670)

### DIFF
--- a/pkg/ovssubnet/common.go
+++ b/pkg/ovssubnet/common.go
@@ -3,7 +3,6 @@ package ovssubnet
 import (
 	"fmt"
 	"net"
-	"os/exec"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


dcbw:
LGTM

sdodson:
How likely is this to be a problem without this fix? We may pull this into
OSE, I know we've had a few reports of this.

danwinship:
> How likely is this to be a problem without this fix? We may pull this into OSE, I know we've had a few reports of this.

openshift restarts docker as part of its own startup, and both openshift/kubernetes and docker fiddle with iptables rules at startup. So it's a race condition but it will definitely happen now and then.

PR #175, opened 2015-10-08
Fix SDN status hook

pravisankar:


pravisankar:
@danwinship PTAL

plugins/osdn/multitenant/plugins.go
```diff
@@ -69,19 +68,5 @@ func (plugin *MultitenantPlugin) TearDownPod(namespace string, name string, id k
...
-	if err != nil {
-		return nil, err
-	}
-	var podNetworkStatus knetwork.PodNetworkStatus
-	err = json.Unmarshal([]byte(out), &podNetworkStatus)
-	if err != nil {
-		return nil, err
-	}
-	return &podNetworkStatus, nil
+	return &knetwork.PodNetworkStatus{}, nil
```
danwinship: That needs to be "return nil, nil" doesn't it? (look at how it gets used in k8s/pkg/kubelet/dockertools/manager.go)

danwinship:
the rest is good. ironically I just noticed this same problem like 10 minutes ago

dcbw:
LGTM.  Good catch!  Why bother execing when we know what it's going to return :)

PR #176, opened 2015-10-08
Admin command to manage project network

pravisankar:
- Join project networks:
    * oadm pod-network join-projects --to=[project-name] [list of projects] | --selector=[project-labels]
- Unisolate project networks:
    * oadm pod-network unisolate-projects [list of projects] | --selector=[project-labels]
- Isolate project networks: 
    * TBD (some preliminary work done)

- Persist Net namespaces with VNID=0
- Release VNID only if no namespace is using it
- openshift network plugins can handle non pod infra container
- openshift-ovs-multitenant/openshift-ovs-subnet cleanup

### DIFF
--- a/pkg/cmd/admin/network/isolate_projects.go
+++ b/pkg/cmd/admin/network/isolate_projects.go
@@ -1,0 +1,78 @@
+package network
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/spf13/cobra"
+
+	kcmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
+	kerrors "k8s.io/kubernetes/pkg/util/errors"
+
+	"github.com/openshift/openshift-sdn/plugins/osdn/multitenant"
+	"github.com/openshift/origin/pkg/cmd/util/clientcmd"
+)
+
+const (
+	IsolateProjectsNetworkCommandName = "isolate-projects"
+
+	isolateProjectsNetworkLong = `
+Isolate project network
+
+Allows projects to isolate their network from other projects when using the %[1]s network plugin.`
+
+	isolateProjectsNetworkExample = `	// Provide isolation for project p1
+	$ %[1]s <p1>
+
+	// Allow all projects with label name=top-secret to have their own isolated project network
+	$ %[1]s --selector='name=top-secret'`
+)
+
+type IsolateOptions struct {
+	Options *ProjectOptions
+}
+
+func NewCmdIsolateProjectsNetwork(commandName, fullName string, f *clientcmd.Factory, out io.Writer) *cobra.Command {
+	opts := &ProjectOptions{}
+	isolateOp := &IsolateOptions{Options: opts}
+
+	cmd := &cobra.Command{
+		Use:     commandName,
+		Short:   "Isolate project network",
+		Long:    fmt.Sprintf(isolateProjectsNetworkLong, multitenant.NetworkPluginName()),
+		Example: fmt.Sprintf(isolateProjectsNetworkExample, fullName),
+		Run: func(c *cobra.Command, args []string) {
+			if err := opts.Complete(f, c, args, out); err != nil {
+				kcmdutil.CheckErr(err)
+			}
+			opts.CheckSelector = c.Flag("selector").Changed
+			if err := opts.Validate(); err != nil {
+				kcmdutil.CheckErr(kcmdutil.UsageError(c, err.Error()))
+			}
+
+			err := isolateOp.Run()
+			kcmdutil.CheckErr(err)
+		},
+	}
+	flags := cmd.Flags()
+
+	// Common optional params
+	flags.StringVar(&opts.Selector, "selector", "", "Label selector to filter projects. Either pass one/more projects as arguments or use this project selector")
+
+	return cmd
+}
+
+func (i *IsolateOptions) Run() error {
+	projects, err := i.Options.GetProjects()
+	if err != nil {
+		return err
+	}
+
+	errList := []error{}
+	for _, project := range projects {
+		// TBD: Create or Update network namespace
+		// TODO: Fix this once we move VNID allocation to REST layer
+		errList = append(errList, fmt.Errorf("Project '%s' can not be isolated. Isolate project network feature yet to be implemented!", project.ObjectMeta.Name))
+	}
+	return kerrors.NewAggregate(errList)
+}
--- a/pkg/ovssubnet/common.go
+++ b/pkg/ovssubnet/common.go
@@ -21,6 +21,8 @@ import (
 const (
 	// Maximum VXLAN Network Identifier as per RFC#7348
 	MaxVNID = ((1 << 24) - 1)
+	// VNID for the admin namespaces
+	AdminVNID = uint(0)
 )
 
 type OvsController struct {
@@ -39,10 +41,14 @@ type OvsController struct {
 
 type FlowController interface {
 	Setup(localSubnetCIDR, clusterNetworkCIDR, serviceNetworkCIDR string, mtu uint) error
+
 	AddOFRules(nodeIP, nodeSubnetCIDR, localIP string) error
 	DelOFRules(nodeIP, localIP string) error
+
 	AddServiceOFRules(netID uint, IP string, protocol api.ServiceProtocol, port uint) error
 	DelServiceOFRules(netID uint, IP string, protocol api.ServiceProtocol, port uint) error
+
+	UpdatePod(namespace, podName, containerID string, netID uint) error
 }
 
 func NewKubeController(sub api.SubnetRegistry, hostname string, selfIP string, ready chan struct{}) (*OvsController, error) {
@@ -83,6 +89,11 @@ func NewController(sub api.SubnetRegistry, hostname string, selfIP string, ready
 	}, nil
 }
 
+func (oc *OvsController) isMultitenant() bool {
+	_, is_mt := oc.flowController.(*multitenant.FlowController)
+	return is_mt
+}
+
 func (oc *OvsController) StartMaster(clusterNetworkCIDR string, clusterBitsPerSubnet uint, serviceNetworkCIDR string) error {
 	subrange := make([]string, 0)
 	subnets, _, err := oc.subnetRegistry.GetSubnets()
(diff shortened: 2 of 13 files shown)


pravisankar:
@danwinship @dcbw @marun  PTAL

pkg/cmd/admin/network/isolate_project.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/openshift/origin/pkg/cmd/util/clientcmd"
+)
+
+const (
+	IsolateProjectNetworkCommandName = "isolate-projects"
+
+	isolateProjectNetworkLong = `
+Isolate project network
+
+Allows project to isolate their network from other projects.`
```
danwinship: Needs to clarify that this only works with the multitenant plugin. ie, it only undoes "unisolate-projects".

(Inconsistency between "project" and "projects"; filename and most variables use the singular, but the actual option uses plural.)

pkg/cmd/admin/network/join_project.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/openshift/origin/pkg/cmd/util/clientcmd"
+)
+
+const (
+	JoinProjectNetworkCommandName = "join-projects"
+
+	joinProjectNetworkLong = `
+Join project network
+
+Allow projects to join existing project network.`
```
danwinship: again probably need to mention multitenant plugin

pkg/cmd/admin/network/join_project.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	projects, err := j.Options.GetProjects()
+	if err != nil {
+		return err
+	}
+
+	errList := []error{}
+	for _, project := range projects {
+		if netID == ovssubnet.VNID0 {
+			err = j.Options.DeleteNetNamespace(project.ObjectMeta.Name)
```
danwinship: won't that cause ovssubnet/common.go to just allocate a new VNID for it on next startup?

pkg/cmd/admin/network/unisolate_project.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/spf13/cobra"
+
+	kcmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
+	kerrors "k8s.io/kubernetes/pkg/util/errors"
+
+	"github.com/openshift/origin/pkg/cmd/util/clientcmd"
+)
+
+const (
+	UnIsolateProjectNetworkCommandName = "unisolate-projects"
```
danwinship: I still don't like "unisolate" as the name for this operation, since I think it only implies "everyone can access this project", not "this project can access everyone". Something with "admin" in the name would be better IMHO

pkg/ovssubnet/common.go
```diff
@@ -21,6 +21,8 @@ import (
 const (
 	// Maximum VXLAN Network Identifier as per RFC#7348
 	MaxVNID = ((1 << 24) - 1)
+	// VNID for the admin namespaces
+	VNID0 = uint(0)
```
danwinship: "AdminVNID"? If we think it might ever be something other than 0, then "VNID0" is a bad name, and if we don't think it might ever be non-0, then there's no reason to not just hardcode 0...

pkg/ovssubnet/controller/kube/bin/openshift-ovs-subnet
```diff
@@ -34,13 +34,16 @@ Setup() {
...
     cluster_subnet=${OPENSHIFT_CLUSTER_SUBNET}
 
-    pid=$(docker inspect --format "{{.State.Pid}}" ${net_container})
     network_mode=$(docker inspect --format "{{.HostConfig.NetworkMode}}" ${net_container})
     if [ "${network_mode}" == "host" ]; then
       # quit, nothing for the SDN here
       exit 0
+    elif [[ "${network_mode}" =~ container:.* ]]; then
+      # Get pod infra container
+      net_container=$(echo ${network_mode} | cut -d ":" -f 2)
```
danwinship: why is this needed? shouldn't all containers in the pod show the same IP address?

dcbw:
Instead of "unisolate" could we do "split" or "leave"?

danwinship:
That's not what unisolate does. Unisolate means "set VNID to 0".
(If you want to split two previously-merged projects A and B, the only way to do that in this UI is to create a new project C, join B to C, and then delete C.)

dcbw:
Yeah, looking at the patches I see there's also an unimplemented "isolate" command.  Which is what I was missing.

Still, seems odd that there isn't a corresponding opposite to Join, like Leave or Split...  If Isolate will eventually create a new VNID for a project/namespace, and Unisolate moves it back into the admin/VNID0 namespace, would the corresponding Leave/Split instead just be calling Isolate on the currently-joined project?

danwinship:
what about:

join (change VNID of PROJECTS[1:] to VNID of PROJECTS[0])
`oadm pod-network join PROJECTS...`
split/isolate: (assign a distinct VNID to each project in PROJECTS)
`oadm pod-network isolate PROJECTS...`
unisolate: (assign VNID 0 to each project in PROJECTS)
`oadm pod-network admin PROJECTS...`

pkg/ovssubnet/controller/kube/bin/openshift-ovs-subnet
pravisankar: If we do nsenter into the container, then yes, all the containers will show the same IP address but docker inspect on non infra pod container won't show the IP addr instead it's NetworkMode points to the pod infra container. I tried to avoid nsenter + parsing in favor of docker inspect on the pod infra container.

pkg/cmd/admin/network/join_project.go
pravisankar: Good catch, do you foresee any issues if we store VNID=0 in the NetNamespace registry? I don't understand why we did not persist namespaces with VNID=0 in the first place.

pkg/cmd/admin/network/isolate_project.go
pravisankar: I'll fix this one. I need to check how we can get network plugin name in pod-network command, that way we can provide useful error message when this cmd is used for flat sdn.

dcbw:
> what about:
> 
> join (change VNID of PROJECTS[1:] to VNID of PROJECTS[0])
> oadm pod-network join PROJECTS...
> split/isolate: (assign a distinct VNID to each project in PROJECTS)
> oadm pod-network isolate PROJECTS...
> unisolate: (assign VNID 0 to each project in PROJECTS)
> oadm pod-network admin PROJECTS...

That looks a bit clearer to me.  What do you think @pravisankar ?

pkg/cmd/admin/network/join_project.go
danwinship: I'm not sure why we didn't either.

pravisankar:
We had some discussion on this before in the email chain, let me recap:
We are creating overlay network for the pods and 'pod-network' was chosen to avoid any confusion with node network.
Eventually 'pod-network' command will support many sub commands once we have granular network policy. 
'-projects' suffix was added to indicate these sub-commands are operated on the projects.
I think, join-projects/isolate-projects provides clear semantics. For join projects, --to option is explicitly used to eliminate any confusion on which project to join, first one or the last one and this pattern is used in other openshift cmds.
When we have more than 2 projects sharing the pod network, split operation might be confusing. splitting from given project or existing pod network? I think, 'isolate' word is less confusing!
For VNID 0 projects:  'unisolate' was chosen to imply 'no isolation' and the help message clears any confusion. 'admin'/'global' are generic, they have other connotations. 'admin network' also needs explanation and this is completely different from the admin we use in ACLs. We need better alternatives if we are not happy with 'unisolate'.

pravisankar:
Updated, PTAL

pkg/cmd/admin/network/isolate_projects.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/openshift/origin/pkg/cmd/util/clientcmd"
+)
+
+const (
+	IsolateProjectsNetworkCommandName = "isolate-projects"
+
+	isolateProjectsNetworkLong = `
+Isolate project network
+
+Allows projects to isolate their network from other projects in case of %[1]s network plugin.`
```
danwinship: I think "Allows projects to isolate their network from other projects (when using the %[1]s network plugin)." would be clearer.

danwinship:
> When we have more than 2 projects sharing the pod network, split operation might be confusing. splitting from given project or existing pod network? I think, 'isolate' word is less confusing!

I think in the case where you pass more than one project, it's not obvious whether that means they all become isolated from each other (each gets its own new VNID), or they all become isolated from everyone else (they all get the same new VNID). The example suggests it's the latter, but I would have guessed the former, since having them share the same VNID means "isolate" also does a "join".

(At any rate, that code isn't hooked up yet, so it doesn't matter that much today.)

> For VNID 0 projects: 'unisolate' was chosen to imply 'no isolation' and the help message clears any confusion.

I think that violates the ordinary meaning of "isolation" though. If project A is described as being "isolated", and project B as "not isolated", no one would expect that to mean that B has full access to A. Because if B has full access to A, then A isn't really isolated. That's why I was suggesting something like "admin"; then we can say that A really is isolated, it's just that B has superpowers and can bypass the isolation. (Although I guess "admin" has the reverse problem; it implies "I can access everyone else", but doesn't imply "everyone else can access me"...)

Let's get the code in at least. Presumably we can still potentially rename it post-dev-cut

dcbw:
Agreed, I had no real problems with the code, just the UX.

pravisankar:
On Mon, Oct 12, 2015 at 9:07 AM, Dan Winship <notifications@github.com>
wrote:

> When we have more than 2 projects sharing the pod network, split operation
> might be confusing. splitting from given project or existing pod network? I
> think, 'isolate' word is less confusing!
>
> I think in the case where you pass more than one project, it's not obvious
> whether that means they all become isolated from each other (each gets its
> own new VNID), or they all become isolated from everyone else (they all get
> the same new VNID). The example suggests it's the latter, but I would have
> guessed the former, since having them share the same VNID means "isolate"
> also does a "join".
>
> (At any rate, that code isn't hooked up yet, so it doesn't matter that
> much today.)
>
> For VNID 0 projects: 'unisolate' was chosen to imply 'no isolation' and
> the help message clears any confusion.
>
> I think that violates the ordinary meaning of "isolation" though. If
> project A is described as being "isolated", and project B as "not
> isolated", no one would expect that to mean that B has full access to A.
> Because if B has full access to A, then A isn't really isolated. That's why
> I was suggesting something like "admin"; then we can say that A really is
> isolated, it's just that B has superpowers and can bypass the isolation.
> (Although I guess "admin" has the reverse problem; it implies "I can access
> everyone else", but doesn't imply "everyone else can access me"...)
>
Instead of admin, may be we can use 'global' (making the project global
within the cluster). Something like 'make-global-projects'?

> Let's get the code in at least. Presumably we can still potentially rename
> it post-dev-cut
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/openshift-sdn/pull/176#issuecomment-147447697>
> .
>



-- 
-Ravi

PR #178, opened 2015-10-09
Service endpoint isolation

danwinship:
Problem: because we only handle service isolation on the source end, you can work around isolation (of both pods and services) by creating a headless service and then manually assigning it endpoints in another project

Solution 1: Use pure-iptables-based service proxying, then do isolation at the destination rather than the source. But this turns out to not work because we still have to masquerade the packets (or else the service will try to send its response directly back to the client, which will reject them because it doesn't have an open connection to the service pod IP) so the destination doesn't know the source IP.

Solution 2: Use OVS-based service proxying instead of iptables so we can carry the source VNID through the whole process. But we still need to masquerade, for the same reason as with iptables, so this can't be done without OVS conntrack support. (Though I think this is the eventual best solution.)

Solution 3: Find some other way to pass the VNID through kube-proxy... Like... maybe kube-proxy would bind its outgoing connection to a source port that had some mathematical relation to the VNID and the destination would check that? Or not.

Solution 3a: use IPv6 addresses and encode information into some of the excess bits. Except we can't use IPv6.

Solution 4: Filter the illegal endpoints out of the list passed to Proxier so we can guarantee that source-side filtering will be sufficient.

This branch implements solution 4. The origin side is at https://github.com/danwinship/origin/commits/endpoint-isolation.

The osdnapi.Pod stuff is going to conflict with Ravi's branch...

@pravisankar @dcbw

### DIFF
--- a/plugins/osdn/osdn.go
+++ b/plugins/osdn/osdn.go
@@ -31,8 +31,8 @@ type OsdnRegistryInterface struct {
 	kClient kclient.Interface
 }
 
-func NewOsdnRegistryInterface(osClient *osclient.Client, kClient *kclient.Client) OsdnRegistryInterface {
-	return OsdnRegistryInterface{osClient, kClient}
+func NewOsdnRegistryInterface(osClient *osclient.Client, kClient *kclient.Client) *OsdnRegistryInterface {
+	return &OsdnRegistryInterface{osClient, kClient}
 }
 
 func (oi *OsdnRegistryInterface) GetSubnets() ([]osdnapi.Subnet, string, error) {
--- a/plugins/osdn/osdn.go
+++ b/plugins/osdn/osdn.go
@@ -2,6 +2,8 @@ package osdn
 
 import (
 	"fmt"
+	"github.com/golang/glog"
+	"net"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


danwinship:
> The osdnapi.Pod stuff is going to conflict with Ravi's branch...

i've just repushed this rebased on top of that

plugins/osdn/osdn.go
```diff
@@ -92,6 +114,56 @@ func (oi *OsdnRegistryInterface) WatchSubnets(receiver chan<- *osdnapi.SubnetEve
...
+	checkCondition := true
+	for {
+		eventType, obj, err := getEvent(eventQueue, startVersion, &checkCondition)
+		if err != nil {
+			return err
+		}
+		kPod := obj.(*kapi.Pod)
+
+		switch eventType {
+		case watch.Added:
```
pravisankar: Don't we need this 'case watch.Added, watch.Modified:'? 
Not sure when is Pod IP set, after creating the pod object or as part of pod creation.

pravisankar:
Rest of the changes LGTM

plugins/osdn/osdn.go
danwinship: oops, yeah. (well, the code works as is, so I guess that means the IP is always added after initial creation)

PR #181, opened 2015-10-13
Detect SDN setup requirement when user switches between flat and multitenant plugins

pravisankar:
This will eliminate migration script.

### DIFF
--- a/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
+++ b/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
@@ -87,6 +87,9 @@ function setup_required() {
     if ! docker_network_config check; then
         return 0
     fi
+    if ! ovs-ofctl -O OpenFlow13 dump-flows br0 | grep -q 'table=0.*arp'; then
+        return 0
+    fi
     return 1
 }
 
--- a/pkg/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
+++ b/pkg/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
@@ -87,6 +87,9 @@ function setup_required() {
     if ! docker_network_config check; then
         return 0
     fi
+    if ! ovs-ofctl -O OpenFlow13 dump-flows br0 | grep -q NXM_NX_TUN_IPV4; then
+        return 0
+    fi
     return 1
 }
 


pravisankar:
@danwinship @dcbw PTAL

pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
```diff
@@ -87,6 +87,9 @@ function setup_required() {
     if ! docker_network_config check; then
         return 0
     fi
+    if ovs-ofctl -O OpenFlow13 dump-flows br0 | grep -q NXM_NX_TUN_IPV4; then
+        return 0
```
danwinship: This will give the wrong result if br0 doesn't exist. The multitenant check doesn't have that problem because of how it uses a "!". Something like:

    if ! ovs-ofctl -O OpenFlow13 dump-flows br0 | grep -q 'table=0.*arp'; then
        return 0
    fi

should work.

(OK, fine, the script *already* does the wrong thing if lbr0 exists but br0 doesn't, but we might as well fix that while we're here...)

brenton:
I apologize if this has been explained elsewhere but could someone describe the behavior admins should expect when switching between the flat and multitenant plugins?  Specifically, what happens to projects that already exist before switching plugins?  Are they automatically switched over or do they keep using their previous network?

danwinship:
The first time you start the master with the multitenant plugin, it will assign a unique network ID to each project. So you'll go from fully-connected to fully-isolated.

https://github.com/openshift/origin/pull/5024 adds "oadm pod-network unisolate ...", which takes a --selector arg, so you could do `oadm pod-network unisolate --selector=*` (or whatever the actual syntax for selectors is) to flip things back.

brenton:
Thanks @danwinship

pravisankar:
Once admin user switches from flat sdn to multitenant sdn, all the existing projects will be full-isolated (just like a new project gets isolated when multitenant network plugin is used). Tuning the pod network of the projects can be done later by 'oadm pod-network'.

pravisankar:
I'll update these details in the openshift-docs

sdodson:
Thanks, we should detail what happens to running pods. They stay in the existing flat network until a node is restarted at which point the pods are restarted into their isolated networks?

pravisankar:
@danwinship  Updated, ptal

danwinship:
oh, I thought we had agreed on IRC that you weren't going to change it? Whatever, either way is fine. lgtm

pravisankar:
When switching from flat to multitenant plugin or vice versa, there is no issue with my change or the change you recommended. However, on a fresh install there won't be any br0 bridge and flat sdn setup won't get executed with my old change.

PR #182, opened 2015-10-13
Don't use vxlan for local subnet communication [DO NOT MERGE]

danwinship:
For containers on other nodes on the same subnet, we can use the nodes as routers and thus forward the traffic unencapsulated.

Proof of concept: not yet supported in multitenant, because it requires more work there (we can't pass the source VNID along, so we'd have to know the VNID of every IP address on every host. Or at least every non-local IP address... Anyway, possible scaling issues...

### DIFF
--- a/pkg/netutils/common.go
+++ b/pkg/netutils/common.go
@@ -4,6 +4,8 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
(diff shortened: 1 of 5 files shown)


danwinship:
Closing for now; this was just a proof-of-concept. We can push a new one when this actually happens.

PR #183, opened 2015-10-16
Fix non-multitenant pod routing

danwinship:
The non-multitenant plugin got broken as part of the changes that went with the new oadm commands. The multitenant plugin still worked fine so none of us noticed... Gotta get those automated tests working...

Fixes https://github.com/openshift/origin/issues/5152

cc @pravisankar @sdodson


pravisankar:
lgtm

sdodson:
Confirmed it fixes the issue for me.

PR #184, opened 2015-10-16
Misc debug.sh fixes

danwinship:
- Parse output of 'host' correctly when /etc/hosts is empty
  - Handle the master being a node
  - Bail out sooner in case of some errors
  - Use $HOME/.kube/config as default KUBECONFIG value

These make it work better/at all with the default OSE config


pravisankar:
lgtm

PR #186, opened 2015-10-20
Clean up stale OVS flows correctly / don't error out of TearDown early

danwinship:
OpenFlow rules were getting left behind because the TearDown was erroring out early because it was trying to remove the veth from the bridge after the veth had already been destroyed.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1272295
@pravisankar


pravisankar:
LGTM

PR #187, opened 2015-10-20
Isolate docker bridge initialization to enable containerized installs

sdodson:
For use in an origin-node image that's built atop openshift/origin. The systemd service that starts the node container is Restart=always and is responsible for starting the node again after docker has been restarted.

### DIFF
--- a/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
+++ b/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
@@ -41,10 +41,12 @@ function docker_network_config() {
 
 DOCKER_NETWORK_OPTIONS='${DOCKER_NETWORK_OPTIONS}'
 EOF
-
-		systemctl daemon-reload
-		systemctl restart docker.service
-
+		## linux bridge
+		ip link set lbr0 down || true
+		brctl delbr lbr0 || true
+		brctl addbr lbr0
+		ip addr add ${local_subnet_gateway}/${local_subnet_mask_len} dev lbr0
+		ip link set lbr0 up
 
 	    if [ ! -f /.dockerinit ]; then
 		# disable iptables for lbr0
@@ -56,6 +58,11 @@ EOF
 		modprobe br_netfilter || true
 		sysctl -w net.bridge.bridge-nf-call-iptables=0
 	    fi
+		# when using --pid=host to run docker container, systemctl inside it refuses
+		# to work because it detects that it's running in chroot. using dbus instead
+		# of systemctl is just a workaround
+		dbus-send --system --print-reply --reply-timeout=2000 --type=method_call --dest=org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager.Reload
+		dbus-send --system --print-reply --reply-timeout=2000 --type=method_call --dest=org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager.RestartUnit string:'docker.service' string:'replace'
 	    ;;
     esac
 }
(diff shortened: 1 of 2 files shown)


sdodson:
This is currently calling setup when all of the checks inside of setup_required() return 0 I'm not sure why that's happening.

https://gist.githubusercontent.com/sdodson/6271d65ec72b3fc3be13/raw/18b5f870af02e4dd7f20cb7a68ef6c5e79c3a59e/gistfile1.txt

danwinship:
> This is currently calling setup when all of the checks inside of setup_required() return 0 I'm not sure why that's happening.

setup_required returns 0 to indicate that setup *is* required, because that value is a process exit status, not a boolean, so it has to be this way for the code later on to make sense:

    if ! setup_required; then
        echo "SDN setup not required."

sdodson:
@danwinship and @marun 

This seems to be the minimum amount of change necessary to run reliably in a container. I tried looking at #166 but it failed to compile so that led me to believe it's probably not something we'd consider including in Origin 1.1 release.

Do you think it's possible to get this in 1.1? My goal is to deliver containerized installs that can run on RHEL Atomic for Origin 1.1 / OSE 3.1.

pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
```diff
@@ -62,10 +64,14 @@ function docker_network_config() {
 
 DOCKER_NETWORK_OPTIONS='${DOCKER_NETWORK_OPTIONS}'
 EOF
-
-		systemctl daemon-reload
-		systemctl restart docker.service
-
+		## linux bridge
+		ip link set lbr0 down || true
```
danwinship: Why did you have to move this here? (And not remove it from setup(); this means lbr0 will be created, then destroyed, then created again.)

Note that the comment "disable iptables for lbr0" just below this section is inaccurate; it actually disables iptables for all bridges, so it ought to work whether or not lbr0 exists yet. (If it doesn't then maybe it's because we need to modprobe something else?)

sdodson: Sorry, I missed removing it from setup(). I moved it here because creating lbr0 and configuring its ip is the minimum we need to have in place to restart docker.

pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
```diff
@@ -62,9 +64,13 @@ function docker_network_config() {
...
-
-		systemctl daemon-reload
-		systemctl restart docker.service
+		## linux bridge
+		ip link set lbr0 down || true
+		brctl delbr lbr0 || true
+		brctl addbr lbr0
+		ip addr add ${local_subnet_gateway}/${local_subnet_mask_len} dev lbr0
+		ip link set lbr0 up
+		brctl addif lbr0 vlinuxbr
```
sdodson: Perhaps this should be moved back to setup() ?

dcbw: > @danwinship Perhaps line 73 should be moved back to setup() as that's where vlinuxbr is created

@sdodson yes, since the docker network config update is done before setup(), which is where vlinuxbr is actually created.

sdodson:
Rebased and squashed commits. Does this seem like a workable solution until the setup is re-written in golang?

pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
```diff
@@ -147,11 +141,15 @@ function setup() {
     echo "export OPENSHIFT_CLUSTER_SUBNET=${cluster_network_cidr}" >> "/etc/openshift-sdn/config.env"
 
     # delete unnecessary routes
-    delete_local_subnet_route lbr0
+    delete_local_subnet_route lbr0 || true
```
sdodson: @pravisankar I'm not sure why, but when running the SDN from within a container this route never shows up so I've added this to tolerate it's non-existence.

danwinship: Hm... that means we'll always have a delay of 2 seconds here while it waits for the route to show up. We should figure out exactly what's up with that at some point.

pravisankar: We should *not* tolerate deletion of lbr0 route.
If we allow creating the lbr0 route then we will run into this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1275388
I thought 2 secs timeout was sufficient, I will do further investigation.

PR #188, opened 2015-10-20
Add firewall-traversing iptables rules even if firewalld isn't running

danwinship:
People may have non-firewalld-based firewall rules. (In particular, our ansible install sets up such rules.)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1273659


pravisankar:
LGTM

PR #190, opened 2015-10-22
Fix multitenant flow cleanup again

danwinship:
Kubernetes now seems to destroy the veths before running Teardown, so it's no longer possible to find the OVS port a container was attached to, so we have to use different logic to clean up after them.

(Last time I stupidly tested against the kube plugin so didn't notice I hadn't fully fixed the multitenant plugin...)

@pravisankar

### DIFF
--- a/pkg/ovssubnet/controller/multitenant/bin/openshift-ovs-multitenant
+++ b/pkg/ovssubnet/controller/multitenant/bin/openshift-ovs-multitenant
@@ -54,19 +54,17 @@ del_ovs_port() {
 add_ovs_flows() {
     ovs_port=$(ovs-ofctl -O OpenFlow13 dump-ports-desc br0  | grep ${veth_host} | cut -d "(" -f 1 | tr -d ' ')
 
-    ovs-ofctl -O OpenFlow13 add-flow br0 "table=3,cookie=0x${ovs_port},priority=100,in_port=${ovs_port},ip,nw_src=${ipaddr},actions=load:${tenant_id}->NXM_NX_REG0[],goto_table:4"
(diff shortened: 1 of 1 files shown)


pkg/ovssubnet/controller/multitenant/bin/openshift-ovs-multitenant
```diff
@@ -54,19 +54,17 @@ del_ovs_port() {
 add_ovs_flows() {
     ovs_port=$(ovs-ofctl -O OpenFlow13 dump-ports-desc br0  | grep ${veth_host} | cut -d "(" -f 1 | tr -d ' ')
 
-    ovs-ofctl -O OpenFlow13 add-flow br0 "table=3,cookie=0x${ovs_port},priority=100,in_port=${ovs_port},ip,nw_src=${ipaddr},actions=load:${tenant_id}->NXM_NX_REG0[],goto_table:4"
+    ovs-ofctl -O OpenFlow13 add-flow br0 "table=3,priority=100,in_port=${ovs_port},ip,nw_src=${ipaddr},actions=load:${tenant_id}->NXM_NX_REG0[],goto_table:4"
```
pravisankar: This will work but can we create a cookie based on service IP? That will be consistent will AddOFRules and kube plugin where cookie is generated based on node IP.

danwinship: We could, but (other than consistency with AddOFRules) is there any reason to? A cookie-based-on-IP would not be any more unique than the IP address itself, so deleting-by-cookie wouldn't be any different from deleting-by-IP.

pravisankar: I looked at the code to know why did we generate cookie based on IP in other places. It was a necessity in other places as there are some OF rules where IP was not present in the rule and cookie based on IP acted as the reference to club related rules for a pod. In the current case, that's not needed. Service IP is already part of the table 3/6 rule. I'm okay with the current change.

PR #191, opened 2015-10-22
debug.sh fixes

danwinship:
Resolves a bunch of problems I've seen in the last week. In particular, the "try to find node config file from 'ps' output" commit is needed to make debug.sh work in our ansible OSE installs.

### DIFF
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -29,6 +29,11 @@ try_eval () {
     return $status
 }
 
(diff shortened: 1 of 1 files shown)


pravisankar:
LGTM

danwinship:
updated https://github.com/danwinship/openshift-sdn/commit/f5c5f738dba22095b9724980f593e2b0f317de4c to completely filter out the "env" section

PR #192, opened 2015-10-22
Remove support for supervisord-managed docker

marun:
docker-in-docker cluster deployment is now managed by systemd, so
supporting supervisord when running in a container is no longer necessary.

### DIFF
--- a/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
+++ b/pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
@@ -25,39 +25,18 @@ function docker_network_config() {
 	DOCKER_NETWORK_OPTIONS="-b=lbr0 --mtu=${mtu}"
     fi
 
+    local conf=/run/openshift-sdn/docker-network
     case "$1" in
 	check)
-	    if [ -f /.dockerinit ]; then
-		# Assume supervisord-managed docker for docker-in-docker deployments
-		conf=/etc/supervisord.conf
-		if ! grep -q -s "DOCKER_DAEMON_ARGS=\"${DOCKER_NETWORK_OPTIONS}\"" $conf; then
-		    return 1
-		fi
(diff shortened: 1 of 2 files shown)


marun:
This is preliminary - support for supervisord still needs to be removed from the multitenant plugin, and the this needs to land in origin at the same time that the dind deployment switches to systemd.

pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
```diff
@@ -25,39 +25,18 @@ function docker_network_config() {
...
-		fi
-
-		echo "Docker networking options have changed; manual restart required." >&2
-		sed -i.bak -e \
-		    "s+\(DOCKER_DAEMON_ARGS=\)\"\"+\1\"${DOCKER_NETWORK_OPTIONS}\"+" \
-		    $conf
-	    else
 		mkdir -p /run/openshift-sdn
-		cat <<EOF > /run/openshift-sdn/docker-network
+		cat <<EOF > $conf
```
danwinship: not sure we really need $conf still, but if you're keeping it, the mkdir should be `mkdir -p $(dirname $conf)`

danwinship:
Basically looks right to me. Need to figure out how this interacts with #187

sdodson:
I can rebase #187 on top of this work

pkg/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh
marun: Done

marun:
Latest patchset updates the multitenant plugin.

danwinship:
lgtm. not merging per your earlier comment about syncing with origin changes

marun:
@danwinship I think this can probably merge now.  The dind move to systemd is almost done and will need this to merge here first.

Issue #151, opened 2015-09-10
remove redundant tun0 route


danwinship:
> ip route
    default via 192.168.0.1 dev eth0 
    10.1.0.0/16 dev tun0  proto kernel  scope link 
    10.1.1.0/24 dev tun0  proto kernel  scope link  src 10.1.1.1 

Rajat says "We should optimize that. The more specific /24 rule gets created because we use ip-tools to configure the ip address on tun0 (which is an internal port to OVS). We will move to go-netlink eventually.. so far no harm though as Brett explained."

PR #193, opened 2015-10-22
Remove redundant tun0 route

pravisankar:
closes: #151

### DIFF
--- a/pkg/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
+++ b/pkg/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
@@ -93,6 +93,28 @@ function setup_required() {
     return 1
 }
 
+# Delete the subnet routing entry created because of ip link up on device
+# ip link adds local subnet route entry asynchronously
+# So check for the new route entry every 100 ms upto timeout of 2 secs and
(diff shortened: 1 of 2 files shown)


pravisankar:
@danwinship ptal

pkg/ovssubnet/controller/multitenant/bin/openshift-sdn-multitenant-setup.sh
```diff
@@ -176,9 +176,12 @@ function setup() {
     sysctl -w net.ipv4.ip_forward=1
     sysctl -w net.ipv4.conf.${TUN}.forwarding=1
 
-    # delete the subnet routing entry created because of lbr0
+    # delete the subnet routing entry created because of ip addr on lbr0
     ip route del ${local_subnet_cidr} dev lbr0 proto kernel scope link src ${local_subnet_gateway} || true
 
+    # delete the subnet routing entry created because of ip addr on ${TUN}
+    ip route del ${local_subnet_cidr} dev ${TUN} proto kernel scope link src ${local_subnet_gateway} || true
```
danwinship: i think it would make more sense to do this immediately after adding the ${cluster_network_cidr} route on tun0

Also, kube plugin?

pravisankar:
@danwinship updated

Issue #194, opened 2015-10-22
Validate when assigning new cluster/service network to the cluster


pravisankar:
Fix this use case:
- Create cluster with default cluster/service network
- Some nodes added to the cluster
- Some services created on the cluster
- Delete default clusterNetwork resource
- Restart openshift-master with new cluster/service network

Existing nodes/services will still use old cluster/service network and new nodes/services will get new cluster/service network. We need to ensure not to run into this mess.

PR #195, opened 2015-10-23
Validate SDN cluster/service network

pravisankar:
closes: #194


pravisankar:
@danwinship @dcbw PTAL

PR #196, opened 2015-10-26
Tolerate error for tun0 redundant route but not for lbr0 route

pravisankar:


pravisankar:
@danwinship @dcbw PTAL

danwinship:
I think it would be more idiomatic to have delete_local_subnet_route() just return 0 or 1, and then later do

    delete_local_subnet_route lbr0
    delete_local_subnet_route tun0 || true

though then you don't get "Error" vs "Warning" in the log message. And this is all eventually going away and being replaced by go code hopefully anyway so it doesn't really matter.

lgtm

pravisankar:
updated

danwinship:
saying "Error/Warning" is silly. Just pick one or the other and then merge it

PR #198, opened 2015-10-27
Refactor plugins

dcbw:
The "subnet" and "vnid" operations need to be split out from common.go so that the upcoming flannel plugin can use "vnid" but ignore "subnet".  This branch also grabs some of Dan Winship's cleanups to the registry.


dcbw:
> Pod watching is needed. We maintain pod IP <--> Namespace mapping (registry.namespaceOfPodIP) and this information is used for filtering service endpoints in case of multitenant solution.

Yeah, you're right.  I didn't see that there was logic in the registry for this until later.  It just seemed odd that the plugins would be doing the watching, but then doing nothing with the result; but the result was being handled by the registry.

dcbw:
@pravisankar @danwinship PTAL

dcbw:
@pravisankar @danwinship PTAL; updated with consolidation of plugins/osdn and pkg/ovssubnet and simplification of the Origin interfaces to openshift-sdn.

danwinship:
(typo in the `Genericize watchAndGetResource()` commit message: "...so watchAdnGetResource() shouldn't...")

danwinship:
generally lgtm

dcbw:
@danwinship @pravisankar PTAL

pravisankar:
LGTM

dcbw:
@danwinship @pravisankar @rajatchopra PTAL

dcbw:
One thing I'd note is that I don't really like the startMasterFuncs/startNodeFuncs stuff, but that actually gets removed in the last commit.  It's only necessary in the interim because until the last commit, plugins are not subclasses of OvsController.

knobunc:
LGTM

rajatchopra:
LGTM

PR #200, opened 2015-10-28
Allow traffic from VNID 0 pods to all services

danwinship:
We had the rules for "all pods can connect to VNID 0 services", but not for "VNID 0 pods can connect to all services".

Closes https://bugzilla.redhat.com/show_bug.cgi?id=1275953


dcbw:
LGTM

pravisankar:
lgtm

PR #201, opened 2015-10-29
Track changes to services

danwinship:
Closes https://bugzilla.redhat.com/show_bug.cgi?id=1275954

One problem; it appears that any time any service changes, we get a modify for *every* service! But modifying services should be rare-ish, so I'd say go with this for 3.1.0 and we can add diffing to get the minimal set of deletes+adds for 3.1.1.

@pravisankar @dcbw

### DIFF
--- a/pkg/ovssubnet/common.go
+++ b/pkg/ovssubnet/common.go
@@ -38,6 +38,7 @@ type OvsController struct {
 	VNIDMap         map[string]uint
 	netIDManager    *netutils.NetIDAllocator
 	AdminNamespaces []string
+	services        map[string]api.Service
 }
 
 type FlowController interface {
@@ -87,6 +88,7 @@ func NewController(sub api.SubnetRegistry, hostname string, selfIP string, ready
 		sig:             make(chan struct{}),
 		ready:           ready,
 		AdminNamespaces: make([]string, 0),
+		services:        make(map[string]api.Service),
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


pravisankar:
LGTM

pkg/ovssubnet/common.go
```diff
@@ -597,9 +604,25 @@ func (oc *OvsController) watchServices(ready chan<- bool, start <-chan string) {
...
+				for _, port := range ev.Service.Ports {
+					oc.flowController.AddServiceOFRules(netid, ev.Service.IP, port.Protocol, port.Port)
+				}
 			case api.Deleted:
-				oc.flowController.DelServiceOFRules(netid, ev.Service.IP, ev.Service.Protocol, ev.Service.Port)
+				delete(oc.services, ev.Service.UID)
+				for _, port := range ev.Service.Ports {
+					oc.flowController.DelServiceOFRules(netid, ev.Service.IP, port.Protocol, port.Port)
+				}
+			case api.Modified:
```
pravisankar: We can do additional check here. If old service IP/protocol/ports match with new service then ignore updating OF rules. This means service update is not relevant to us, update related to SessionAffinity/ExternalIPs/Selector or some other fields.

pkg/ovssubnet/common.go
```diff
@@ -597,9 +604,38 @@ func (oc *OvsController) watchServices(ready chan<- bool, start <-chan string) {
...
+				oldsvc, exists := oc.services[ev.Service.UID]
+				if exists && len(oldsvc.Ports) == len(ev.Service.Ports) {
+					same := true
+					for i := range oldsvc.Ports {
+						if oldsvc.Ports[i].Protocol != ev.Service.Ports[i].Protocol || oldsvc.Ports[i].Port != ev.Service.Ports[i].Port {
+							same = false
+							break
+						}
+					}
+					if same {
```
dcbw: Would reflect.DeepEqual() on Ports be easier here?

dcbw:
LGTM though

dcbw:
[merge]

PR #202, opened 2015-10-29
debug.sh: fix handling of pods with no ip

danwinship:
See https://bugzilla.redhat.com/show_bug.cgi?id=1274591#c5

The errors here make me realize that we're operating on potentially-untrusted user-generated data from within a shell script that needs to be run as root... We're going to need to fix that. Until then, we should remember that the script is not safe for OpenShift users who don't trust all of their own users.


pravisankar:
lgtm

PR #203, opened 2015-10-30
Detect host network conflict with sdn cluster/service network

pravisankar:


pravisankar:
@danwinship @dcbw  PTAL

pkg/ovssubnet/common.go
```diff
@@ -137,11 +149,16 @@ func (oc *OvsController) validateServiceNetwork(networkCIDR string) error {
 }
 
 func (oc *OvsController) validateNetworkConfig(clusterNetworkCIDR, serviceNetworkCIDR string, subnetsInUse []string) error {
+	hostIPs, err := net.LookupIP(oc.hostName)
```
danwinship: Would it be better to loop over `net.Interfaces()`? That would let us check multiple interfaces, and would get all IPs on the system, rather than just the ones listed in DNS.

pravisankar:
Updated, ptal

danwinship:
lgtm

PR #206, opened 2015-11-04
Fix netutils tests

danwinship:
openshift-sdn's (minimal) tests fail now because pkg/netutils/common.go now requires kubernetes's pkg/util/errors. So this just copies that into Godeps. I didn't bother trying to do it "properly". Maybe we should just rm Godeps/Godeps.json and Godeps/Readme?

@pravisankar

### DIFF
--- a/Godeps/_workspace/src/k8s.io/kubernetes/pkg/util/errors/errors.go
+++ b/Godeps/_workspace/src/k8s.io/kubernetes/pkg/util/errors/errors.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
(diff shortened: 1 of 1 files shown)


pravisankar:
Other option is to move GetHostIPNetworks() to pkg/ovssubnet/common.go (used only in this file). 
I'm okay with either copying kerrors file or moving the method.
Currently, hack/sync-to-origin.sh checks Godeps/Godeps.json to confirm the current repo belongs to openshift-sdn. We need to patch the script if we plan to get rid of Godeps.json.

danwinship:
> Other option is to move GetHostIPNetworks() to pkg/ovssubnet/common.go

I'm still assuming we're eventually moving all this to origin, so functions should stay where they make the most sense

PR #207, opened 2015-11-06
Remove unused lookup of VNID for TearDownPod

dcbw:
The script doesn't care about the VNID, so we shouldn't bother
looking it up only to fail if the teardown request comes after
the master has sent the NetNamespace deleted notification.

https://bugzilla.redhat.com/show_bug.cgi?id=1277383

### DIFF
--- a/plugins/osdn/multitenant/plugins.go
+++ b/plugins/osdn/multitenant/plugins.go
@@ -52,11 +52,8 @@ func (plugin *MultitenantPlugin) SetUpPod(namespace string, name string, id kube
 }
 
 func (plugin *MultitenantPlugin) TearDownPod(namespace string, name string, id kubeletTypes.DockerID) error {
-	vnid, found := plugin.OvsController.VNIDMap[namespace]
-	if !found {
-		return fmt.Errorf("Error fetching VNID for namespace: %s", namespace)
-	}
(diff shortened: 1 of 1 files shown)


plugins/osdn/multitenant/plugins.go
```diff
@@ -52,11 +52,7 @@ func (plugin *MultitenantPlugin) SetUpPod(namespace string, name string, id kube
 }
 
 func (plugin *MultitenantPlugin) TearDownPod(namespace string, name string, id kubeletTypes.DockerID) error {
-	vnid, found := plugin.OvsController.VNIDMap[namespace]
-	if !found {
-		return fmt.Errorf("Error fetching VNID for namespace: %s", namespace)
-	}
-	out, err := utilexec.New().Command(plugin.getExecutable(), tearDownCmd, namespace, name, string(id), strconv.FormatUint(uint64(vnid), 10)).CombinedOutput()
+	out, err := utilexec.New().Command(plugin.getExecutable(), tearDownCmd, namespace, name, string(id), "-1").CombinedOutput()
```
rajatchopra: We should probably put a comment here about the executable not caring about vnid.

pravisankar:
LGTM

plugins/osdn/multitenant/plugins.go
dcbw: > We should probably put a comment here about the executable not caring about vnid.

Done.

PR #208, opened 2015-11-12
Merge flatsdn and multitenant code

danwinship:
The flatsdn/"kube"/"subnet" and multitenant code have always been nearly identical so let's finally merge all the duplicate code (while still retaining the two plugin names with their different behavior).

(Not yet tested... want to make sure that this makes sense first)

@dcbw, @pravisankar


danwinship:
test of what happens if I try to tag @kubernetes/rh-network...

danwinship:
(tested now)

plugins/osdn/ovs/controller.go
```diff
@@ -0,0 +1,161 @@
...
+		arprule = fmt.Sprintf("table=0,cookie=0x%s,priority=100,arp,nw_dst=%s,actions=set_field:%s->tun_dst,output:1", cookie, nodeSubnetCIDR, nodeIP)
+	}
+	out, err := exec.Command("ovs-ofctl", "-O", "OpenFlow13", "add-flow", "br0", iprule).CombinedOutput()
+	if err != nil {
+		glog.Errorf("Error adding flow '%s': %s (%v)", iprule, out, err)
+		return err
+	}
+	out, err = exec.Command("ovs-ofctl", "-O", "OpenFlow13", "add-flow", "br0", arprule).CombinedOutput()
+	if err != nil {
+		glog.Errorf("Error adding flow '%s': %s (%v)", iprule, out, err)
```
pravisankar: *arprule

plugins/osdn/ovs/controller.go
```diff
@@ -0,0 +1,161 @@
...
+	if nodeIP == localIP {
+		return nil
+	}
+
+	glog.V(5).Infof("DelOFRules for %s", nodeIP)
+
+	rule := fmt.Sprintf("cookie=0x%s/0xffffffff", generateCookie(nodeIP))
+	out, err := exec.Command("ovs-ofctl", "-O", "OpenFlow13", "del-flows", "br0", rule).CombinedOutput()
+	if err != nil {
+		glog.Errorf("Error deleting flow '%s': %s (%v)", rule, out, err)
```
pravisankar: nit: %q for quoted string('%s')

pravisankar:
minor nitpicks, otherwise LGTM
We also need to update node install/vagrant provision/rpm spec scripts in the origin repo and
sdn.adoc in openshift-docs repo.

danwinship:
> We also need to update node install/vagrant provision/rpm spec scripts in the origin repo and
> sdn.adoc in openshift-docs repo.

It looks like sdn.adoc is non-specific enough that it doesn't need any updates.

danwinship:
@openshift/networking I think this is ready to go in?

dcbw:
In the middle of testing it actually, but it all looks OK to me so far.

dcbw:
ok, so we have a problem with the setup script where, if /run/openshift-sdn/docker-network is correct but lbr0 doesn't exist, the node service will error out.  We'll need to update that, but I think we should merge this first and then update that after.

In any case, after I sorted out the setup issue, light testing to make sure things got set up correctly with each plugin went OK.

PR #209, opened 2015-11-16
Shuffle things around in node start to fix a hang

danwinship:
See https://github.com/openshift/origin/pull/5840 (specifically https://github.com/openshift/origin/pull/5840/files#r44958452). We need to return the FilteringEndpointsConfigHandler sooner, so that StartNode() can be called as a goroutine, because StartNode() can't return until after its caller does, because of the weird way startup is currently organized.

(origin side is https://github.com/danwinship/origin/tree/osdn-update)

@dcbw

### DIFF
--- a/plugins/osdn/api/types.go
+++ b/plugins/osdn/api/types.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"github.com/openshift/origin/pkg/cmd/server/kubernetes"
 	knetwork "k8s.io/kubernetes/pkg/kubelet/network"
 )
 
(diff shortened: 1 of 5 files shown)


pravisankar:
I think StartNode() was called as a goroutine so that later on RunKubelet() can do the node registration with master which will assign the subnet to the node. Is there an issue if we call RunKubelet() first and then call RunSDNController() (without calling StartNode() as a go routine)?

pravisankar:
Code changes LGTM but not sure if this is the right approach...It depends on the answer to my question raised in previous comment.

pravisankar:
We have a dependency between RunSDNController() and RunKubelet() on KubeletConfig.StartUpdates channel. We have a trello card for this issue: https://trello.com/c/o7LT58cW . Until we fix that card, for now we can run StartNode() as a go routine.

pravisankar:
Origin changes (https://github.com/openshift/origin/compare/master...danwinship:osdn-update?expand=1) LGTM

PR #210, opened 2015-11-18
[DO_NOT_MERGE] Refactor VNID allocator

pravisankar:
- Added admin CLI command to isolate projects


pravisankar:
@openshift/networking PTAL

plugins/osdn/vnidallocator/allocator.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+type Allocator struct {
+	vnidRange util.VNIDRange
+
+	alloc allocator.Interface
+}
+
+// Allocator implements Interface and Snapshottable
+var _ Interface = &Allocator{}
```
danwinship: what does this do? It looks like it just creates and then throws away an Allocator

plugins/osdn/vnidallocator/allocator.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Free returns the count of VNID left in the range.
+func (r *Allocator) Free() int {
+	return r.alloc.Free()
+}
+
+// Allocate attempts to reserve the provided VNID. ErrNotInRange or
+// ErrAllocated will be returned if the VNID is not valid for this range
+// or has already been reserved. ErrFull will be returned if there
+// are no VNIDs left.
```
danwinship: Either (a) vnid is out of range, in which case ErrNotInRange, or (b) it's in range but already used, in which case ErrAllocated, or (c) it's in range and not already used, in which case Allocate() will succeed. It shouldn't be possible to get ErrFull.

plugins/osdn/vnidallocator/allocator.go
```diff
@@ -0,0 +1,146 @@
...
+	return r.vnidRange.Base + uint(offset), nil
+}
+
+// Release releases the VNID back to the pool. Releasing an
+// unallocated VNID or a VNID out of the range is a no-op and
+// returns no error.
+func (r *Allocator) Release(vnid uint) error {
+	ok, offset := r.contains(vnid)
+	if !ok {
+		glog.Warningf("Tried to release vnid: %d which is unallocated or out of the range", vnid)
```
danwinship: It's weird to say that this "is a no-op and returns no error" if you're going to log a warning. Either silently allow it, or document it as being not allowed.

plugins/osdn/vnids.go
```diff
@@ -42,101 +17,24 @@ func (oc *OvsController) VnidStartMaster() error {
...
-		return err
-	}
-	oc.VNIDMap[namespaceName] = netid
-	return nil
-}
 
-func (oc *OvsController) revokeVNID(namespaceName string) error {
-	err := oc.Registry.DeleteNetNamespace(namespaceName)
+	// Handle existing namespaces without corresponding netnamespaces
+	netNamespaceMap := make(map[string]bool, len(namespaces))
```
danwinship: should be len(netnsList) really

danwinship:
Is there a corresponding origin branch? It seems like this removes a bunch of code and doesn't replace it with anything...

plugins/osdn/vnidallocator/allocator.go
pravisankar: On the origin side, NetNamespace REST layer will use vnidallocator.Interface

plugins/osdn/vnidallocator/allocator.go
pravisankar: Yes, I'll update

pravisankar:
Updated

plugins/osdn/vnidallocator/allocator.go
danwinship: Sorry, github made my comment unclear. I was only referring to the final line:

    +var _ Interface = &Allocator{}

why do you create and then throw away an Allocator object? How is this not a no-op?

pravisankar: This is a common practice. It will ensure Allocator struct has implemented all the methods declared in k8s allocater.Interface, otherwise it will throw an error like "*Allocator does not implement Interface (missing XYZ method)".

pravisankar:
@openshift/ui-review  commit 3baa342 adds CLI command to support network isolation for projects (undo merged project networks).

fabianofranz:
CLI bits LGTM

pravisankar:
@openshift/networking Updated, ptal

plugins/osdn/factory/factory.go
```diff
@@ -24,3 +24,10 @@ func NewPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Cl
 
 	return nil, nil, nil
 }
+
+func IsMultitenantNetworkPlugin(pluginType string) bool {
```
danwinship: Would it make things simpler to just always do the VNID allocation stuff, even when using ovs-subnet? (The code would just ignore the VNIDs when creating OVS flows in that case.)

danwinship:
vnidallocator doesn't get used by anything else in openshift-sdn... could it be moved to origin instead? Then all the code would be in one place, which seems like a win.

plugins/osdn/factory/factory.go
pravisankar: Enabling NetNamespace REST resource for ovs-subnet is not an issue but the VNID repair controller runs every 15 mins and makes a bunch of etcd calls and these will return nothing in case of ovs-subnet. So enabling NetNamespace/VNID only when it is necessary will save some cpu/io cycles.

pravisankar:
Yes, vnidallocator is only used in origin. It makes sense to move the code to origin.

dcbw:
I'm wondering how this would all work out when splitting all the plugin stuff out of openshift[-sdn].  The VNID allocator would go into our future Kube network provider, I guess?

pravisankar:
Moved VNID allocator related code to origin. This also eliminated unnecessary copy of some k8s files under Godeps which were needed for the test cases.

pravisankar:
@danwinship @dcbw rebased, ptal

plugins/osdn/factory/factory.go
```diff
@@ -23,3 +23,10 @@ func NewPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Cl
...
+	if strings.ToLower(pluginType) == ovs.MultiTenantPluginName() {
```
dcbw: This part is going to be fun when flannel bits land, since those need the VNID stuff too, but I'll just deal with that when things settle down.

dcbw:
LGTM assuming the origin bits get OKd.

pravisankar:
Rebased, resolved minor conflicts.

danwinship:
LGTM. Fine to merge once someone approves the origin side to go. (ie, let's not land this until we know it's not going to need more changes to address comments on the origin PR)

pravisankar:
Closing this PR in favor of https://github.com/openshift/openshift-sdn/pull/303

PR #212, opened 2015-11-19
Don't return an error if the plugin is unknown

dcbw:
We need to silently accept unknown plugins in origin, so if we don't know about it, don't return an error that Origin will exit on.


pravisankar:
If there is a typo in the network plugin name, this will hide the issue and returns nil.
Instead I prefer the caller/origin to do something like:
if nodeConfig.NetworkConfig.NetworkPluginName != "" {
       controller, endpointFilter, err := factory.NewPlugin(...)
      ...
}

dcbw:
I know, but the problem is that we need to support existing OpenShift plugins that are not openshift-sdn.  Like Contrail...  Those will specify a network plugin name that we don't match in openshift-sdn, but it will still get carried through to Kubernetes and be used there.

danwinship:
lgtm. @pravisankar we didn't log anything before in that case either, so this is consistent with the old code

PR #217, opened 2015-12-03
[DO NOT MERGE] Add flannelmt network plugin

dcbw:
Pending flannel upstream acceptance of the flannel side, see https://github.com/coreos/flannel/pull/384

Comments/review gratefully accepted though.

### DIFF
--- a/plugins/osdn/flannelmt/bin/openshift-sdn-flannelmt
+++ b/plugins/osdn/flannelmt/bin/openshift-sdn-flannelmt
@@ -1,0 +1,125 @@
+#!/bin/bash
+set -e
+
+lock_file=/var/lock/openshift-sdn.lock
+
+action=$1
+pod_namespace=$2
+pod_name=$3
+net_container=$4
+tenant_id=$5
+
+lockwrap() {
+    (
+    flock 200
+    "$@"
+    ) 200>${lock_file}
+}
+
+# Retrieve the name of the host-local member of the veth pair that
+# connects the container (identified by pid) to the docker bridge.
+get_veth_host() {
+    local pid=$1
+
+    local veth_ifindex=$(nsenter -n -t $pid -- ethtool -S eth0 | sed -n -e 's/.*peer_ifindex: //p')
+    # Strip a suffix starting with '@' from the interface name.
+    # The suffixed interface name won't be recognized by brctl or ovs-*
+    ip link show | sed -ne "s/^$veth_ifindex: \([^:@]*\).*/\1/p"
+}
+
+get_ipaddr_pid_veth() {
+    network_mode=$(docker inspect --format "{{.HostConfig.NetworkMode}}" ${net_container})
+    if [ "${network_mode}" == "host" ]; then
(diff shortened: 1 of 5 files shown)


danwinship:
I had thought the plan was that the flannel plugin would go directly into origin, with openshift-sdn only being for the (eventually "legacy") openshift-ovs plugins?

plugins/osdn/flannelmt/bin/openshift-sdn-flannelmt
```diff
@@ -0,0 +1,125 @@
...
+lockwrap() {
+    (
+    flock 200
+    "$@"
+    ) 200>${lock_file}
+}
+
+# Retrieve the name of the host-local member of the veth pair that
+# connects the container (identified by pid) to the docker bridge.
+get_veth_host() {
```
danwinship: Should we be abstracting stuff out rather than duplicating code?

plugins/osdn/flannelmt/bin/openshift-sdn-flannelmt
```diff
@@ -0,0 +1,125 @@
...
+}
+
+del_ovs_port() {
+    ovs-vsctl --if-exists del-port $veth_host
+}
+
+add_ovs_flows() {
+    ovs_port=$(ovs-ofctl -O OpenFlow13 dump-ports-desc br0  | grep ${veth_host} | cut -d "(" -f 1 | tr -d ' ')
+
+    ovs-ofctl -O OpenFlow13 add-flow br0 "table=3,priority=100,in_port=${ovs_port},ip,nw_src=${ipaddr},actions=load:${tenant_id}->NXM_NX_REG0[],goto_table:4"
```
danwinship: This isn't going to work; you're making openshift-sdn depend on the precise details of how flannel sets up OVS, so this locks you to a single specific version of flannel. We have a hard enough time keeping openshift-sdn and origin in sync; there's no way we can try to stay in sync with an external module.

Flannel needs to install some script that we can call here that abstracts out these details.

plugins/osdn/vnids.go
```diff
@@ -105,6 +105,7 @@ func (oc *OvsController) assignVNID(namespaceName string) error {
 		return err
 	}
 	oc.VNIDMap[namespaceName] = netid
+
```
danwinship: accidental whitespace change?

dcbw:
> I had thought the plan was that the flannel plugin would go directly into origin, with openshift-sdn only being for the (eventually "legacy") openshift-ovs plugins?

We can do that too, but the current code uses quite a bit of the existing stuff on the openshift-sdn side.  We could copy that to origin and sync periodically, and maybe that makes more sense since Ravi is putting stuff into origin now anyway.  In the end, we still have to have somewhere for these pieces to go even when they are pure Kube and CNI plugins, and origin probably makes sense for that right now.

pravisankar:
Needs rebase

dcbw:
dropping this one for now...

PR #221, opened 2015-12-04
Make kubelet to block on network plugin initialization

pravisankar:


pravisankar:
@openshift/networking PTAL

plugins/osdn/common.go
```diff
@@ -214,10 +214,27 @@ func (oc *OvsController) StartNode(mtu uint) error {
 		return fmt.Errorf("Failed to start plugin: %v", err)
 	}
 
-	if oc.ready != nil {
-		close(oc.ready)
+	oc.MarkPodNetworkReady()
```
dcbw: MarkPodNetworkReady() doesn't need to be accessed from outside the OVSController, right?  Maybe we should lower-case the 'M' then so it's not exported.

plugins/osdn/common.go
```diff
@@ -214,10 +214,27 @@ func (oc *OvsController) StartNode(mtu uint) error {
...
+
+	return nil
+}
+
+func (oc *OvsController) MarkPodNetworkReady() {
+	if oc.podNetworkReady != nil {
```
dcbw: It looks like we're always setting up oc.podNetworkReady, so we don't need to check for nil at all here.

pravisankar:
Updated

danwinship:
So this lets us get rid of the k8s patch, which is good, but it still requires the goroutine goofiness discussed in #209. It would be nice if we could untangle that somehow. (Maybe later.)

danwinship:
lgtm

dcbw:
LGTM

PR #223, opened 2015-12-08
Add warning message when got 127.0.0.1 from Node host

nak3:


nak3:
I sent this PR because a user hit an issue which he got 127.0.0.1 from Node host IP. 
And it was difficult to debug from current error message only.

dcbw:
I'd rather have it check whether len(addrs) == 1 && addrs[0] == "127.0.0.1" and then return a specific error message.  The current patch will print the error no matter what, even if there are other addresses that would be used instead.

pravisankar:
Closing this pr in favor of https://github.com/openshift/openshift-sdn/pull/224

PR #224, opened 2015-12-08
Fix sdn GetHostIPNetworks() and GetNodeIP()

pravisankar:


pravisankar:
@openshift/networking ptal

dcbw:
LGTM

PR #225, opened 2015-12-09
Move pod network readiness check to pod setup instead of network plugin init

pravisankar:


pravisankar:
@dcbw @danwinship PTAL

dcbw:
Moving it here is a bit weird because it would block the first pod started on the node without logging why that's happening, we should at least add some more logging here before starting to wait.  In any case, this solves the problem though I feel like there's gotta be a cleaner solution.  Maybe that waits until the Kube network plugin stuff.

danwinship:
Given the origin patch, could we move the blocking to SubnetStartNode()?

pravisankar:
The scope of the podNetworkReady channel is not specific to SubnetStartNode(), we also need in StartNode() and I prefer this channel to be mutated in one place for sanity.

danwinship:
oh, i thought it only mattered to subnet.go. But anyway, same question: could we move the blocking to StartNode()? (RunKubelet() will already be running asynchronously at that point, so it seems like it should be safe to block there.)

pravisankar:
@danwinship 
Not sure if I understood your question correctly.
No, we don't want to block on StartNode(). StartNode() does the sdn related setup and once it is ready, it will close the podNetworkReady channel. For every new pod, kubelet will call the network plugin setup hook/api  with pod infra details and we don't want to proceed if sdn setup is not ready. So the blocking call has to be in the pod setup hook.

pravisankar:
Added log message when pod is waiting for sdn initialization.

danwinship:
Hm... ok, I was confused about which thread was waiting for what. So the flow is:

- StartNode() (`origin/pkg/cmd/server/start/start_node.go`)
  - RunKubelet() (`start_node.go`)
    - spawns a goroutine to call KubeletServer.Run() (`kubernetes/cmd/kubelet/app/server.go`)
      - RunKubelet() (`server.go`)
        - CreateAndInitKubelet() (`server.go`)
          - NewMainKubelet() (`kubernetes/pkg/kubelet/kubelet.go`)
            - InitNetworkPlugin() (`kubernetes/pkg/kubelet/network/plugins.go`)
              - calls our plugin.Init()
            - spawns a goroutine to call syncNodeStatus() (`kubelet.go`)
              - registerWithApiServer() (`kubelet.go`)
                - calls kClient.Nodes().Create(), which then results in the master creating the subnet record for the node
        - startKubelet() (`server.go`)
          - spawns goroutines
            - eventually receives pod creation requests
            - calls our SetUpPod() which calls WaitForPodNetworkReady()
  - RunSDNController() (`start_node.go`)
    - StartNode() (`openshift-sdn/plugins/osdn/common.go`)
      - PluginStartNode() (`openshift-sdn/plugins/osdn/ovs/plugin.go`)
        - SubnetStartNode() (`openshift-sdn/plugins/osdn/subnets.go`)
          - calls initSelfSubnet() and blocks until the subnet record is created
          - runs `openshift-sdn-ovs-setup.sh`
      - markPodNetworkReady() (`common.go`)

OK. I was thinking that it was the code after StartNode() returns that cared whether the pod network is ready, but it's not, it's the code inside the goroutine spawned by RunKubelet... and plugin.Init() is definitely too early, and there doesn't seem to be any place we can plug in between there and SetUpPod().

OK. LGTM.

danwinship:
> I feel like there's gotta be a cleaner solution. Maybe that waits until the Kube network plugin stuff.

If the call to InitNetworkPlugin() was moved to later in NewMainKubelet(), that would solve the problem (making Ravi's original patch work). This would not be hard to do; the plugin needs to be *created* early so that it can be passed to dockertools.NewDockerManager(), but the code could be reorganized so that it's created early without calling Init(), and then we explicitly call Init() later (after the syncNodeStatus() goroutine, presumably right around the point where the volume plugins get inited).

danwinship:
(The other possibility, I guess, would be to split server.go:RunKubelet() into separate CreateKubelet() and RunKubelet() calls, so we can do the former before calling RunSDNController() and the latter after.)

PR #227, opened 2015-12-16
Subnet allocation fixes

danwinship:
Given ClusterNetworkCIDR "10.1.0.0/16" and HostSubnetLength 7, the current git master algorithm would allocate subnets in the order:

  - 10.1.0.0/25
  - 10.1.0.128/25
  - 10.1.1.0/25
  - 10.1.1.128/25

and then when you see a pod IP like "10.1.0.130", you have to stop and think to figure out which node it's on. Lame!

This branch fixes it so that the allocation order becomes:

  - 10.1.0.0/25
  - 10.1.1.0/25
  - 10.1.2.0/25
  - ...
  - 10.1.255.0/25
  - 10.1.0.128/25
  - 10.1.1.128/25
  - 10.1.2.128/25
  - ...

meaning that if you have fewer than 256 nodes, then every node will have a unique 3rd octet, just like with the HostSubnetLength=8 case.

Oh, and then there's also a mostly-unrelated change to registry.go that removes the redundant validation of ClusterNetwork changes so that https://github.com/openshift/origin/pull/6349 will work.

@openshift/networking

### DIFF
--- a/pkg/netutils/subnet_allocator.go
+++ b/pkg/netutils/subnet_allocator.go
@@ -7,18 +7,18 @@ import (
 
 type SubnetAllocator struct {
 	network  *net.IPNet
-	capacity uint
+	hostBits uint
 	allocMap map[string]bool
 }
 
-func NewSubnetAllocator(network string, capacity uint, inUse []string) (*SubnetAllocator, error) {
+func NewSubnetAllocator(network string, hostBits uint, inUse []string) (*SubnetAllocator, error) {
 	_, netIP, err := net.ParseCIDR(network)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to parse network address: %q", network)
 	}
 
 	netMaskSize, _ := netIP.Mask.Size()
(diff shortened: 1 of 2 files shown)


danwinship:
> Oh, and then there's also a mostly-unrelated change to registry.go that removes the redundant validation of ClusterNetwork changes so that openshift/origin#6349 will work.

split out into its own PR. Now this is just the subnet numbering fix

danwinship:
> t.Fatalf

fixed

>      	hostBits uint 
> can we make this uint32?

It's not a 32-bit number, it's a number from 1 to 31, so plain `uint` seems more appropriate. (I renamed it from "capacity" to try to make it clearer that it's "the number of bits in the address that are reserved for the host", but maybe "hostBits" isn't the best name?)

Likewise with leftShift/rightShift.

danwinship:
@pravisankar can you re-review?

eparis:
[merge]

eparis:
wrong tree  :)

PR #228, opened 2015-12-16
Move FilteringEndpointsConfigHandler into openshift-sdn

dcbw:
It's currently osdn specific anyway, plus we need it here to avoid
an import loop once some of liggitt's suggestions are taken for
origin SDN initialization ordering.


dcbw:
See https://github.com/dcbw/origin/pull/1/files for the general direction here for origin.

dcbw:
@pravisankar @danwinship @rajatchopra

pravisankar:
LGTM

PR #229, opened 2015-12-18
Don't redundantly validate the ClusterNetwork

danwinship:
Split out of #227. This is the openshift-sdn side of https://github.com/openshift/origin/pull/6349.

@openshift/networking


pravisankar:
lgtm

PR #230, opened 2015-12-18
Bump node subnet wait to 30 seconds (from 10)

dcbw:
The node subnet wait is run from a goroutine and will run
concurrently with the rest of kubelet setup.  But when a node
is heavily loaded kubelet setup might actually take close to
10 seconds or more before registration with the master starts.
That causes the subnet wait to time waiting for the subnet
registration which hasn't happened yet.  Bump the timeout.

https://bugzilla.redhat.com/show_bug.cgi?id=1290967


danwinship:
lgtm

PR #233, opened 2015-12-24
Admin commands should not depend on OVS code

smarterclayton:
Admin commands are client centric, and should not take dependencies on
plugin implementations (which brings in lots of undesirable packages).
This change will enable the admin part of the origin CLI to be embedded
into oc.


pravisankar:
LGTM

PR #235, opened 2016-01-05
Make pkg/ipcmd and pkg/ovs not panic if their binaries are missing

danwinship:
Fixes the breakage seen in https://github.com/openshift/origin/pull/6532

The openshift binary gets run as part of the tests/CI, so we can't panic if OVS is missing. But just returning an error will still cause it to exit at startup time when running normally, so that's OK. (And
it looks like nothing else caches the result of LookPath(), so we won't either; this is necessary to make it possible to test the missing-command codepaths.)

@dcbw

### DIFF
--- a/pkg/ipcmd/ipcmd.go
+++ b/pkg/ipcmd/ipcmd.go
@@ -2,23 +2,16 @@
 package ipcmd
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
 
 	"github.com/openshift/openshift-sdn/pkg/exec"
 )
 
-var ipcmdPath string
 var addressRegexp *regexp.Regexp
 
 func init() {
-	var err error
-
-	ipcmdPath, err = exec.LookPath("ip")
-	if err != nil {
-		panic("ip is not installed")
-	}
-
 	addressRegexp = regexp.MustCompile("inet ([0-9.]*/[0-9]*) ")
 }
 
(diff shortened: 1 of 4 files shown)


pkg/ovs/ovs.go
```diff
@@ -40,18 +25,24 @@ func (tx *Transaction) exec(cmd string, args ...string) (string, error) {
 		return "", tx.err
 	}
 
+	cmdpath, err := exec.LookPath(cmd)
```
pravisankar: We are following this pattern:
  Start Transaction
  Bunch of operations (calls exec to perform the op)
  End Transaction
Instead of doing LookPath(cmd) for every exec(), another option is to set vsctlPath and ofctlPath in NewTransaction() if not initialized. Any operation is performed using transaction object, so this will ensure vsctl/ofctl paths always set.

dcbw:
LGTM; no longer panics on startup in my testing.

dcbw:
@pravisankar thanks GitHub for not auto-updating with your comments before I hit merge, I should have refreshed first...  want @danwinship to do an update?

pkg/ipcmd/ipcmd.go
```diff
@@ -2,23 +2,16 @@
...
 	"github.com/openshift/openshift-sdn/pkg/exec"
 )
 
-var ipcmdPath string
 var addressRegexp *regexp.Regexp
 
 func init() {
-	var err error
-
-	ipcmdPath, err = exec.LookPath("ip")
```
pravisankar: don't we expect 'ip' binary to be present? we do call 'ip link/route ...' in few places. May be for non-redhat network plugin, we can't  expect this binary.
Another option is to set ipcmdPath in NewTransaction() similar to ovs command path.

danwinship: We *expect* it, yes, but we can't guarantee it in all cases.

Returning an error here means that openshift won't start up if it's not present, which is good

pkg/ovs/ovs.go
danwinship: Yeah, we could do that. It looks like everywhere else in openshift/kubernetes doesn't really worry about caching path lookups though...

PR #237, opened 2016-01-06
Add a route to fix up service IP usage from the node.

### LINKED ISSUE #232: openshift: cannot access (remote) service using portal address from node
Using the multitenant plugin on atomic hosts (using openvswitch on docker), openshift v1.1, I cannot access a service portal address (172.30.183.139:5000) on the host.

> curl 172.30.183.139:5000

times out.

> curl 172.30.183.139:5000 --interface tun0

works.

as does using the endpoint (10.1.2.13) directly.

When adding a route to the portal net, all requests are working.

> ip r add 172.30.0.0/16 dev tun0

The relevant part of the (generated) nat table looks like that.

```
-A PREROUTING -m comment --comment "kubernetes service portals" -j KUBE-SERVICES
-A OUTPUT -m comment --comment "kubernetes service portals" -j KUBE-SERVICES
-A POSTROUTING -s 10.1.0.0/16 ! -d 10.1.0.0/16 -j MASQUERADE
-A POSTROUTING -m comment --comment "kubernetes service traffic requiring SNAT" -m mark --mark 0x4d415351 -j MASQUERADE
-A KUBE-SEP-FIQ6M4FASARUD5ON -s 10.1.2.13/32 -m comment --comment "default/docker-registry:5000-tcp" -j MARK --set-xmark 0x4d415351/0xffffffff
-A KUBE-SEP-FIQ6M4FASARUD5ON -p tcp -m comment --comment "default/docker-registry:5000-tcp" -m tcp -j DNAT --to-destination 10.1.2.13:5000
-A KUBE-SVC-ECTPRXTXBM34L34Q -m comment --comment "default/docker-registry:5000-tcp" -j KUBE-SEP-FIQ6M4FASARUD5ON
-A KUBE-SERVICES -d 172.30.183.139/32 -p tcp -m comment --comment "default/docker-registry:5000-tcp cluster IP" -m tcp --dport 5000 -j KUBE-SVC-ECTPRXTXBM34L34Q

```

danwinship:
(This was broken by the switch to iptables-based services, because the request would use the default IP rather than the tun0 IP, and then the response would come back via the wrong path and not get un-NAT-ed.)

Closes #232 

@dcbw: maybe get this into the resync too?

### DIFF
--- a/plugins/osdn/ovs/controller.go
+++ b/plugins/osdn/ovs/controller.go
@@ -242,6 +242,7 @@ func (c *FlowController) Setup(localSubnetCIDR, clusterNetworkCIDR, servicesNetw
 	defer deleteLocalSubnetRoute(TUN, localSubnetCIDR)
 	itx.SetLink("up")
 	itx.AddRoute(clusterNetworkCIDR, "proto", "kernel", "scope", "link")
+	itx.AddRoute(servicesNetworkCIDR)
 	err = itx.EndTransaction()
 	if err != nil {
 		return err


dcbw:
LGTM

Issue #240, opened 2016-01-11
debug.sh should test node/pod connectivity


danwinship:
We test pod-to-pod connectivity,but not node-to-pod / pod-to-node

PR #243, opened 2016-01-12
debug.sh fixups

danwinship:
This started because I was going to implement #240, but then the bug I needed that for ended up fixed anyway, so I haven't bothered yet. But here are some cleanups and improvements.

I bet @knobunc wants to review this...

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1265586

### DIFF
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -492,8 +492,14 @@ do_master_and_nodes ()
 	echo ""
 	echo "Analyzing $name ($addr)"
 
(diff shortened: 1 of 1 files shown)


dcbw:
Could you add one more bit to this?  In the node-specific data, can you grab

sysctl -a | grep bridge

to make sure that bridge-nf-call-iptables is correct.

danwinship:
Done. (Well, actually I just output all of "sysctl -a".)

dcbw:
LGTM

PR #244, opened 2016-01-12
use net.bridge.bridge-nf-call-iptables sysctl key

ibotty:
I don't know whether that's correct, but I figure that bridge-nf-call is a typo. I did not have a kernel in reach that had _that_ sysctl. When using net.bridge.bridge-nf-call-iptables, container egress works alright.

Note, that I did not get into the trouble to verify that change by building a new openshift/node image.


dcbw:
Yeah, this is clearly a typo and is a regression introduced in d48ee2f28991ae0eaf0c92be31659bd66b85f7c4

dcbw:
Can confirm this fixes the issue seen on openshift-114.lab.

PR #245, opened 2016-01-12
Ensure NodeIP doesn't overlap with the cluster network

dcbw:
If a HostSubnet is edited or the node's IP address is changed such
that the NodeIP is contained within its own SubnetCIDR, a routing loop
will be created in the OVS bridge and cause a kernel panic by
exhausing kernel stack space.

Prevent the cluster from starting when this happens by ensuring that
the NodeIP is not contained within the cluster network CIDR.

https://bugzilla.redhat.com/show_bug.cgi?id=1295486

### DIFF
--- a/plugins/osdn/common.go
+++ b/plugins/osdn/common.go
@@ -169,7 +169,7 @@ func (oc *OvsController) StartMaster(clusterNetworkCIDR string, clusterBitsPerSu
 	// Any mismatch in cluster/service network is handled by WriteNetworkConfig
 	// For any new cluster/service network, ensure existing node subnets belong
 	// to the given cluster network and service IPs belong to the given service network
-	if _, err := oc.Registry.GetClusterNetworkCIDR(); err != nil {
+	if _, err := oc.Registry.GetClusterNetwork(); err != nil {
 		subrange := make([]string, 0)
 		subnets, _, err := oc.Registry.GetSubnets()
 		if err != nil {
@@ -178,6 +178,11 @@ func (oc *OvsController) StartMaster(clusterNetworkCIDR string, clusterBitsPerSu
 		}
 		for _, sub := range subnets {
 			subrange = append(subrange, sub.SubnetCIDR)
+		}
+
+		_, _, _, err = ValidateClusterNetwork(clusterNetworkCIDR, int(clusterBitsPerSubnet), serviceNetworkCIDR)
+		if err != nil {
+			return err
 		}
 
 		err = oc.validateNetworkConfig(clusterNetworkCIDR, serviceNetworkCIDR, subrange)
@@ -199,19 +204,19 @@ func (oc *OvsController) StartMaster(clusterNetworkCIDR string, clusterBitsPerSu
 
 func (oc *OvsController) StartNode(mtu uint) error {
 	// Assume we are working with IPv4
-	clusterNetworkCIDR, err := oc.Registry.GetClusterNetworkCIDR()
+	clusterNetwork, err := oc.Registry.GetClusterNetwork()
 	if err != nil {
 		log.Errorf("Failed to obtain ClusterNetwork: %v", err)
 		return err
 	}
 
 	ipt := iptables.New(kexec.New(), utildbus.New(), iptables.ProtocolIpv4)
-	if err := SetupIptables(ipt, clusterNetworkCIDR); err != nil {
+	if err := SetupIptables(ipt, clusterNetwork.String()); err != nil {
 		return fmt.Errorf("Failed to set up iptables: %v", err)
 	}
 
 	ipt.AddReloadFunc(func() {
-		err := SetupIptables(ipt, clusterNetworkCIDR)
+		err := SetupIptables(ipt, clusterNetwork.String())
 		if err != nil {
 			log.Errorf("Error reloading iptables: %v\n", err)
 		}
--- a/plugins/osdn/common.go
+++ b/plugins/osdn/common.go
@@ -90,64 +90,7 @@ func (oc *OvsController) BaseInit(registry *Registry, flowController FlowControl
 	return nil
 }
 
-func (oc *OvsController) validateClusterNetwork(networkCIDR string, subnetsInUse []string, hostIPNets []*net.IPNet) error {
-	clusterIP, clusterIPNet, err := net.ParseCIDR(networkCIDR)
-	if err != nil {
-		return fmt.Errorf("Failed to parse network address: %s", networkCIDR)
-	}
-
-	errList := []error{}
-	for _, ipNet := range hostIPNets {
-		if ipNet.Contains(clusterIP) {
-			errList = append(errList, fmt.Errorf("Error: Cluster IP: %s conflicts with host network: %s", clusterIP.String(), ipNet.String()))
-		}
-		if clusterIPNet.Contains(ipNet.IP) {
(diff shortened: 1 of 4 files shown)


danwinship:
Looks good, although should this be validated by the node rather than the master? I guess if the master fails then the node won't ever be able to finish starting up, so having the master validate it will ensure that the node doesn't create the routing loop anyway.

pravisankar:
Yes, This will handle node IP conflicts with cluster network for exiting subnets but we also need to handle one more case. We allow node IP changes. We watch for this change and when this happens, we recreate the subnet. New subnet creation in this case should be only allowed if the new node IP doesn't conflict with cluster network.

dcbw:
Fixed it up so that node events are now checked for validity too.

plugins/osdn/common.go
```diff
@@ -83,30 +84,49 @@ func (oc *OvsController) BaseInit(registry *Registry, flowController FlowControl
 	return nil
 }
 
-func (oc *OvsController) validateClusterNetwork(networkCIDR string, subnetsInUse []string, hostIPNets []*net.IPNet) error {
-	clusterIP, clusterIPNet, err := net.ParseCIDR(networkCIDR)
+func validateNode(oc *OvsController, nodeIP string) error {
```
marun: Why is this not a method of OvsController?

plugins/osdn/common.go
```diff
@@ -159,26 +179,18 @@ func (oc *OvsController) validateNetworkConfig(clusterNetworkCIDR, serviceNetwor
...
-		subrange := make([]string, 0)
-		subnets, _, err := oc.Registry.GetSubnets()
-		if err != nil {
-			log.Errorf("Error in initializing/fetching subnets: %v", err)
-			return err
-		}
-		for _, sub := range subnets {
-			subrange = append(subrange, sub.SubnetCIDR)
-		}
+	var err error
```
marun: Consider naming the return type instead of declaring a new variable (e.g. func foo() (err error) {}).

danwinship:
LGTM

plugins/osdn/common.go
```diff
@@ -29,6 +29,7 @@ type OvsController struct {
 	Registry        *Registry
 	localIP         string
 	localSubnet     *api.Subnet
+	clusterNet      *net.IPNet
```
marun: It feels a bit strange to me to use a field to pass this argument into the validation methods.  Is StartMaster always guaranteed to be called before the validation methods?

plugins/osdn/subnets.go
```diff
@@ -179,11 +185,15 @@ func watchNodes(oc *OvsController, ready chan<- bool, start <-chan string) {
...
-							continue
+						if nodeErr == nil {
+							sub.NodeIP = ev.Node.IP
+							err = oc.Registry.CreateSubnet(ev.Node.Name, sub)
+							if err != nil {
+								log.Errorf("Error creating subnet for node %s, ip %s", ev.Node.Name, sub.NodeIP)
+								continue
+							}
+						} else {
+							log.Errorf("Deleting invalid node %s/%s subnet: %v", ev.Node.Name, ev.Node.IP, nodeErr)
```
marun: Is there a subsequent action here that I'm missing to differentiate line 177's stated intention to ignore the node vs this line's state intention to delete it?

danwinship:
> should this be validated by the node rather than the master?

dcbw realized later that the answer to this is probably "yes", because if the master exits immediately when the data in etcd is invalid, then you can't use "oc" to fix it.

dcbw:
I chose to do two sides to this.  The master should warn about invalid Node IPs and refuse to create HostSubnet objects for them.  The node should exit if its address is bad, and should ignore invalid HostSubnets from other nodes so we don't create flow rules for them.

plugins/osdn/common.go
danwinship: > Specifically I'm wondering about the order of execution between StartMaster and watchNodes.

watchNodes() only gets called from within SubnetStartMaster(), which gets called from within StartMaster() after oc.clusterNet is set, so this should be ok

plugins/osdn/common.go
```diff
@@ -166,26 +171,11 @@ func (oc *OvsController) validateNetworkConfig(clusterNetworkCIDR, serviceNetwor
...
-
-		err = oc.validateNetworkConfig(clusterNetworkCIDR, serviceNetworkCIDR, subrange)
-		if err != nil {
-			return err
-		}
+	if err := oc.validateNetworkConfig(clusterNetworkCIDR, serviceNetworkCIDR); err != nil {
+		return err
 	}
 
+	// Any mismatch in cluster/service network is handled by WriteNetworkConfig
```
danwinship: It's not really clear what this comment means... You can probably just delete it.

plugins/osdn/common.go
danwinship: Um, that comment applies to the old code, not the current code. But it's ok in the current code too.

But it's weird that the field is only set for masters, not for nodes. And that in some places, you've plumbed clusterNetworkCIDR through (eg, PluginStartNode()), but in other places, it uses the field.

Perhaps we should just make Registry.serviceNetwork and Registry.clusterNetwork public so we can refer to them everywhere? Although then there's the whole set of issues that went with making sure those get set right (https://github.com/openshift/origin/pull/6684) which might apply to oc.clusterNet too?

plugins/osdn/subnets.go
```diff
@@ -96,23 +104,25 @@ func (oc *OvsController) deleteNode(nodeName string) error {
...
+		return false, fmt.Errorf("Failed to parse cluster network address %s: %v", clusterNetworkCIDR, err)
 	}
+
 	servicesNetworkCIDR, err := oc.Registry.GetServicesNetworkCIDR()
 	if err != nil {
 		log.Errorf("Failed to obtain ServicesNetwork: %v", err)
 		return false, err
 	}
+
+	if err := oc.initSelfSubnet(oc.clusterNet); err != nil {
```
danwinship: Why does this get moved? In particular, this is the step that waits for the master to have finished initializing, so moving it to after the GetServicesNetworkCIDR() call might cause that call to fail.

plugins/osdn/subnets.go
```diff
@@ -96,23 +104,25 @@ func (oc *OvsController) deleteNode(nodeName string) error {
...
-	err := oc.initSelfSubnet()
-	if err != nil {
-		return false, err
-	}
+func (oc *OvsController) SubnetStartNode(clusterNetworkCIDR string, mtu uint) (bool, error) {
+	var err error
 
 	// Assume we are working with IPv4
-	clusterNetworkCIDR, err := oc.Registry.GetClusterNetworkCIDR()
+	_, oc.clusterNet, err = net.ParseCIDR(clusterNetworkCIDR)
```
danwinship: why not use GetClusterNetworkCIDR?

dcbw:
@danwinship @marun repushed; now the Registry caches the clusterNetwork so we can get rid of oc.clusterNetwork.

dcbw:
And now broken up into separate commits for easier reviewing...

dcbw:
@pravisankar fixed your comment, PTAL

danwinship:
LGTM. Have you run origin's tests with this code yet?

pravisankar:
Rest of the changes, LGTM

dcbw:
Reworked to address Ravi's comments.

dcbw:
@pravisankar PTAL

knobunc:
LGTM

dcbw:
@danwinship in passes extended tests in DIND at least.

pravisankar:
LGTM

PR #247, opened 2016-01-15
Don't crash on endpoints with invalid/empty IP addresses

danwinship:
Prevents https://bugzilla.redhat.com/show_bug.cgi?id=1298942. Not exactly sure why there is an invalid endpoint, but apparently it only happens while something is being restarted. Anyway, the log message may help figure out what's actually happening.

@openshift/networking


eparis:
@smarterclayton looks like a possible 3.1.1 to me. please review and merge if you agree.

smarterclayton:
Approved, LGTM [merge]

pravisankar:
LGTM

PR #248, opened 2016-01-15
Crash in the right place if the ClusterNetwork record can't be read

danwinship:
This reverts #247 as it turns out that it wasn't actually "IP" that was nil there, it was registry.serviceNetwork.

This new version requires origin changes too, because when creating the plugin, we now need to know whether we're the master or a node.

Not tested yet, so don't merge it. Waiting to hear from Jeremy on https://bugzilla.redhat.com/show_bug.cgi?id=1298942


smarterclayton:
LGTM

PR #250, opened 2016-01-18
endpoints nil crash fix (this time for sure!)

danwinship:
Already tested in origin; jenkins and travis pass.

The origin node startup code calls `StartSDN()` followed by `StartProxy()`. `StartSDN()` calls `oc.StartNode()`, which depends on `ClusterNetwork.Get("default")` returning non-nil (via `registry.GetClusterNetworkCIDR()`), so when `StartProxy()` calls `SetBaseEndpointsHandler()`, we know it will succeed.

(Note that most of this patch is just reverting #248.)

@dcbw PTAL

### DIFF
--- a/plugins/osdn/registry.go
+++ b/plugins/osdn/registry.go
@@ -1,7 +1,6 @@
 package osdn
 
 import (
-	"fmt"
 	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


plugins/osdn/registry.go
```diff
@@ -33,39 +32,19 @@ type Registry struct {
 	oClient osclient.Interface
 	kClient kclient.Interface
 
+	// These are only set if SetBaseEndpointsHandler() has been called
 	baseEndpointsHandler pconfig.EndpointsConfigHandler
 	serviceNetwork       *net.IPNet
 	clusterNetwork       *net.IPNet
 	namespaceOfPodIP     map[string]string
```
pravisankar: May be we want to move this line above the "// These are only set..." comment.

pravisankar:
LGTM

danwinship:
re-pushed, reorganizing the Registry struct declaration per the last comment

pravisankar:
Travis failed on gofmt.
Need gofmt -w plugins/osdn/registry.go

danwinship:
> Travis failed on gofmt

that was the previous branch. it's already fixed in the current version

PR #251, opened 2016-01-18
Ensure bridge-nf-call-iptables is always disabled on startup

dcbw:
The Kubernetes proxy and external entities could set bridge-nf-call-iptables
to 1 so we need to ensure it's still 0 after the node starts up.  In
particular the Kube proxy runs after SDN setup (because it relies on the
endpoint filters that the SDN can provide) so there's no better way to
ensure that bridge-nf-call-iptables stays 0 than a goroutine and loop.

### DIFF
--- a/plugins/osdn/ovs/controller.go
+++ b/plugins/osdn/ovs/controller.go
@@ -6,7 +6,6 @@ import (
 	"github.com/golang/glog"
 	"io/ioutil"
 	"net"
-	"os/exec"
 	"strings"
 	"time"
 
@@ -265,17 +264,6 @@ func (c *FlowController) Setup(localSubnetCIDR, clusterNetworkCIDR, servicesNetw
 	itx.DeleteLink()
 	itx.IgnoreError()
 	_ = itx.EndTransaction()
-
-	// Disable iptables for linux bridges (and in particular lbr0), ignoring errors.
-	// (This has to have been performed in advance for docker-in-docker deployments,
-	// since this will fail there).
-	_, _ = exec.Command("modprobe", "br_netfilter").CombinedOutput()
-	err = sysctl.SetSysctl("net/bridge/bridge-nf-call-iptables", 0)
-	if err != nil {
-		glog.Warningf("Could not set net.bridge.bridge-nf-call-iptables sysctl: %s", err)
-	} else {
-		glog.V(5).Infof("[SDN setup] set net.bridge.bridge-nf-call-iptables to 0")
-	}
 
 	// Enable IP forwarding for ipv4 packets
 	err = sysctl.SetSysctl("net/ipv4/ip_forward", 1)
--- a/plugins/osdn/api/types.go
+++ b/plugins/osdn/api/types.go
@@ -86,6 +86,8 @@ type OsdnPlugin interface {
 
 	StartMaster(clusterNetworkCIDR string, clusterBitsPerSubnet uint, serviceNetworkCIDR string) error
 	StartNode(mtu uint) error
+
+	NodeInitialized()
 }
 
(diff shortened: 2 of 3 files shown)


plugins/osdn/ovs/controller.go
```diff
@@ -142,6 +142,24 @@ func (c *FlowController) Setup(localSubnetCIDR, clusterNetworkCIDR, servicesNetw
...
+	// Hack around the Kube proxier also sets bridge-nf-call-iptables after we do
+	go func() {
+		for i := 0; i < 20; i++ {
+			if val, err := sysctl.GetSysctl("net/bridge/bridge-nf-call-iptables"); val != 0 || err != nil {
+				glog.V(5).Infof("Setting bridge-nf-call-iptables=0 (currently %d)", val)
+				if err := sysctl.SetSysctl("net/bridge/bridge-nf-call-iptables", 0); err != nil {
+					glog.Warningf("Could not set net.bridge.bridge-nf-call-iptables sysctl: %s", err)
+				}
+			}
+			time.Sleep(time.Second)
```
pravisankar: Do we want to sleep for 20 secs if GetSysctl() returns value = 0?

liggitt: I wouldn't consider this a real fix. If we need hooks for pre/post setup notifications or blocking, we should plumb those.

dcbw: > I wouldn't consider this a real fix. If we need hooks for pre/post setup notifications or blocking, we should plumb those.

Fair enough, I'll do those.

dcbw:
@liggitt is this more what you're thinking of?  Associated Origin change is then:

        config.RunSDN()
        config.RunProxy()
 
-       // HACK: RunProxy resets bridge-nf-call-iptables from what openshift-sdn requires
        if config.SDNPlugin != nil {
-               sdnPluginName := nodeConfig.NetworkConfig.NetworkPluginName
-               if sdnPluginName == osdn.SingleTenantPluginName() || sdnPluginName == osdn.MultiTenantPluginName() {
-                       if err := sysctl.SetSysctl("net/bridge/bridge-nf-call-iptables", 0); err != nil {
-                               glog.Warningf("Could not set net.bridge.bridge-nf-call-iptables sysctl: %s", err)
-                       }
-               }
+               config.SDNPlugin.NodeInitialized()
        }
 
        return nil

liggitt:
I'm still trying to understand the core issue. Is this the required order:
1. start sdn plugin (via `StartNode()`)
2. start the proxy (does this depend on things done in the sdn plugin in `StartNode()`?)
3. tell the sdn plugin the proxy is started?

If so, we should change the `StartNode()` interface to take a `proxyReady chan struct{}`, and inside StartNode(), the sdn plugin can do:
```
go func(){
  defer util.HandleCrash()
  <-proxyReady
  ... do the setup ...
}()
```

liggitt:
talked on irc... I think @dcbw is going to check upstream whether it makes sense to optionally prevent the proxy from messing with the sysctl setting (or have the proxy set it to what we want), instead of compensating by having the SDN plugin set it, then fix it up

danwinship:
That kinda ties in to https://github.com/kubernetes/kubernetes/issues/19766, which is about not setting hairpin mode if the plugin doesn't need it

dcbw:
@liggitt @danmcp https://github.com/kubernetes/kubernetes/pull/20647

dcbw:
If that gets accepted, it should be pretty easy to plumb through start_node.go and server.go.

dcbw:
Closing this one in favor of https://github.com/kubernetes/kubernetes/pull/20647

PR #252, opened 2016-01-20
Put the network-already-set-up check in the right place

danwinship:
This ended up in the wrong place after one of the rewrites/rebases. We should check if everything is already set up correctly *before* forcibly recreating lbr0.

This didn't matter until #249 landed, because alreadySetUp() was always returning false. But now, if nothing has changed; lbr0 gets recreated, but then alreadySetUp() returns true before we re-add vlinuxbr to it, so non-OpenShift containers would be broken.

@dcbw PTAL


pravisankar:
LGTM

PR #256, opened 2016-02-05
If NodeIP and NodeName are unsuitable, fallback to default interface …

dcbw:
…IP address

Better match NodeIP determination with upstream kubernetes' setNodeAddress()
function, which eventually falls back to getting the first IP address of
the interface with the default route.

https://bugzilla.redhat.com/show_bug.cgi?id=1304582


pravisankar:
LGTM

PR #257, opened 2016-02-12
Resync from origin for kubernetes rebase

danwinship:
Changes got made to origin's copy of openshift-sdn but not to openshift-sdn itself


danwinship:
@openshift/networking PTAL

danwinship:
oops, just a sec, doesn't build

pravisankar:
lgtm

PR #264, opened 2016-02-19
Re-fix reconnect-pods-on-restart code

danwinship:
When using multi-tenant, we need to run VnidStartNode() before updating the pods, to ensure that VNIDMap is initialized. But VnidStartNode() assumes that SubnetStartNode() has already run. So,
move the pod-updating code out of SubnetStartNode() into PluginStartNode().


danwinship:
@openshift/networking

pravisankar:
LGTM

PR #265, opened 2016-02-29
Fixes to danwinship/arp-fixes

danwinship:
Fixes to arp-fixes based on the testing done in https://bugzilla.redhat.com/show_bug.cgi?id=1311702:

- Add a missing del-flows for new ARP flows on pod deletion.
- Remove the newly-added anti-spoofing checks for tun0 traffic, which break in the face of service-related masquerading.
- Fix the handling of VNID 0 service flows: we need to create the rules for them, just without a reg0 check.

@openshift/networking

### DIFF
--- a/plugins/osdn/ovs/bin/openshift-sdn-ovs
+++ b/plugins/osdn/ovs/bin/openshift-sdn-ovs
@@ -82,6 +82,7 @@ del_ovs_flows() {
     ovs-ofctl -O OpenFlow13 del-flows br0 "ip,nw_dst=${ipaddr}"
     ovs-ofctl -O OpenFlow13 del-flows br0 "ip,nw_src=${ipaddr}"
     ovs-ofctl -O OpenFlow13 del-flows br0 "arp,nw_dst=${ipaddr}"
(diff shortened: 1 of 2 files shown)


dcbw:
> Remove the newly-added anti-spoofing checks for tun0 traffic, which break in the face of service-related masquerading.

What's the nw_src and nw_dst on these packets?

danwinship:
The comment was specifically referring to return traffic from services, which would be in ServicesNetworkCIDR, but also, return traffic from the internet will come in through tun0, so nw_src could be just about anything. I guess we could do:

    table=0, priority=250, in_port=2, ip, nw_src=${localSubnetGateway}, actions=goto_table:5
    table=0, priority=200, in_port=2, ip, nw_src=${clusterNetworkCIDR}, actions=drop
    table=0, priority=150, in_port=2, ip, actions=goto_table:5

but what's the threat model that protects us from?

The nw_dst check was valid, but it doesn't actually add any security, because if nw_dst is outside of clusterNetworkCIDR then OVS is either going to drop the packet, or bounce it back out tun0.

dcbw:
LGTM

PR #266, opened 2016-03-04
Implement OVS bandwidth shaping

dcbw:
Use "standard" kubernetes ingress/egress annotations to implement bandwidth shaping via OVS.

### DIFF
--- a/plugins/osdn/ovs/bin/openshift-sdn-ovs
+++ b/plugins/osdn/ovs/bin/openshift-sdn-ovs
@@ -6,6 +6,9 @@ lock_file=/var/lock/openshift-sdn.lock
 action=$1
 net_container=$2
 tenant_id=$3
+ingress_bw=$4
+egress_bw=$5
+mtu=$6
 
 lockwrap() {
     (
@@ -76,6 +79,19 @@ add_ovs_flows() {
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=0, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=${tenant_id}, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
     fi
+
+    # Pod ingress == OVS bridge egress
+    # linux-htb used here since that's the Kubernetes default traffic shaper too
+    if [ -n "${ingress_bw}" ]; then
+        qos=$(ovs-vsctl create qos type=linux-htb other-config:max-rate=${ingress_bw})
+        ovs-vsctl set port ${veth_host} qos=${qos}
+    fi
+
+    # Pod egress == OVS bridge ingress
+    if [ -n "${egress_bw}" ]; then
+        # OVS ingress_policing specified in Kbps
+        ovs-vsctl set interface ${veth_host} ingress_policing_rate=$((${egress_bw}/1000))
+    fi
 }
 
 del_ovs_flows() {
@@ -83,6 +99,12 @@ del_ovs_flows() {
     ovs-ofctl -O OpenFlow13 del-flows br0 "ip,nw_src=${ipaddr}"
     ovs-ofctl -O OpenFlow13 del-flows br0 "arp,nw_dst=${ipaddr}"
     ovs-ofctl -O OpenFlow13 del-flows br0 "arp,nw_src=${ipaddr}"
+
+    qos=$(ovs-vsctl get port ${veth_host} qos)
+    if [ -n "$qos" ]; then
+        ovs-vsctl clear port ${veth_host} qos
+        ovs-vsctl destroy qos ${qos}
+    fi
 }
 
 add_subnet_route() {
--- a/plugins/osdn/ovs/plugin.go
+++ b/plugins/osdn/ovs/plugin.go
@@ -11,6 +11,7 @@ import (
 	"github.com/openshift/openshift-sdn/plugins/osdn/api"
 
 	kapi "k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/resource"
 	kubeletTypes "k8s.io/kubernetes/pkg/kubelet/container"
 	knetwork "k8s.io/kubernetes/pkg/kubelet/network"
 	utilexec "k8s.io/kubernetes/pkg/util/exec"
@@ -19,6 +20,7 @@ import (
 type ovsPlugin struct {
 	osdn.OvsController
 
+	MTU         uint
 	multitenant bool
 }
 
@@ -60,6 +62,8 @@ func (plugin *ovsPlugin) PluginStartMaster(clusterNetwork *net.IPNet, hostSubnet
 }
 
 func (plugin *ovsPlugin) PluginStartNode(mtu uint) error {
+	plugin.MTU = mtu
+
 	networkChanged, err := plugin.SubnetStartNode(mtu)
 	if err != nil {
 		return err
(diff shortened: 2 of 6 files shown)


plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -76,6 +78,15 @@ add_ovs_flows() {
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=0, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=${tenant_id}, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
     fi
+
+    if [ -n "${ingress_bw}" ]; then
+        ovs-vsctl set port ${veth_host} qos=@newqos -- --id=@newqos create qos type=linux-htb other-config:max-rate=${ingress_bw}
```
danwinship: The new syntax we're not using anywhere else yet is made even more confusing by the fact that `@newqos` gets used before it's set (which is apparently allowed, but... weird). Could you do the `create` first and then the `set`?

Also, something somewhere (commit message?) should document why linux-htb rather than any of the others

plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -76,6 +78,15 @@ add_ovs_flows() {
...
+    fi
+
+    if [ -n "${egress_bw}" ]; then
+        ovs-vsctl set interface ${veth_host} ingress_policing_rate=${egress_bw}
+        ovs-vsctl set interface ${veth_host} ingress_policing_burst=500
```
danwinship: This blatantly looks like a cut+paste error and probably needs a comment clarifying that it isn't: "ingress" here means "ingress into ovs", and thus corresponds to the container's egress.

Also, why 500? And why hardcoded? The man page says "The burst size should be at least the size of the interface’s MTU" and also that "a value that is numerically at least as large as 10% of ingress_policing_rate helps TCP"

plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -83,6 +94,17 @@ del_ovs_flows() {
     ovs-ofctl -O OpenFlow13 del-flows br0 "ip,nw_src=${ipaddr}"
     ovs-ofctl -O OpenFlow13 del-flows br0 "arp,nw_dst=${ipaddr}"
     ovs-ofctl -O OpenFlow13 del-flows br0 "arp,nw_src=${ipaddr}"
+
+    qos=$(ovs-vsctl list port ${veth_host} >&/dev/null | grep qos | cut -d ":" -f 2)
```
danwinship: You probably copied that from somewhere else but there's a cleaner way to do it:

    ovs-vsctl get Port ${veth_host} qos

plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -83,6 +94,17 @@ del_ovs_flows() {
...
+    if [ -n "$qos" ]; then
+        queue=$(ovs-vsctl list qos ${qos} | grep queues | cut -d ":" -f 2 | cut -d "=" -f 2 | cut -d "}" -f 1)
```
danwinship: maybe other people don't like regexes, but I feel like using sed here would be simpler:

    queue=$(ovs-vsctl get QoS ${qos} queues | sed -e 's/.*\([0-9]*\)=.*/\1/')

(not tested since I can't figure out how to make queues end up non-empty)

plugins/osdn/ovs/plugin.go
```diff
@@ -124,25 +125,77 @@ func (plugin *ovsPlugin) getVNID(namespace string) (string, error) {
 	return "0", nil
 }
 
+var minRsrc = resource.MustParse("1k")
+var maxRsrc = resource.MustParse("1P")
+
+func validateBandwidthIsReasonable(rsrc *resource.Quantity) error {
+	if rsrc.Value() < minRsrc.Value() {
+		return fmt.Errorf("resource is unreasonably small (< 1kbit)")
```
danwinship: use minRsrc in the error message? otherwise someone's going to change the min/max and forget to change them here too

plugins/osdn/ovs/plugin.go
```diff
@@ -124,25 +125,77 @@ func (plugin *ovsPlugin) getVNID(namespace string) (string, error) {
...
+	str, found := pod.Annotations["kubernetes.io/ingress-bandwidth"]
+	if found {
+		ingressQuantity, err := resource.ParseQuantity(str)
+		if err != nil {
+			return "", "", err
+		}
+		if err := validateBandwidthIsReasonable(ingressQuantity); err != nil {
+			return "", "", err
+		}
+		ingress = fmt.Sprintf("%d", ingressQuantity.Value())
```
danwinship: it seems wrong to have this function return numbers as strings, just because we happen to want strings later on

plugins/osdn/ovs/bin/openshift-sdn-ovs
danwinship: > Also, something somewhere (commit message?) should document why linux-htb rather than any of the others

or even just a comment in this PR

dcbw:
@danwinship PTAL; I think I've addressed all your comments.  Thanks!

plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -76,6 +79,30 @@ add_ovs_flows() {
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=0, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=${tenant_id}, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
     fi
+
+    # Pod ingress == OVS vSwitch egress
```
danwinship: s/OVS/Open/ ?

plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -76,6 +79,30 @@ add_ovs_flows() {
...
+    fi
+
+    # Pod egress == OVS vSwitch ingress
+    if [ -n "${egress_bw}" ]; then
+        # OVS ingress_policing specified in Kbps
+        egress_kbps=$((${egress_bw}/1000))
+        ovs-vsctl set interface ${veth_host} ingress_policing_rate=${egress_kbps}
+
+        # OVS also recommends that the burst be the greater of the interface
+        # MTU or 1/10th the ingress_policing_rate
```
danwinship: Hm... so I take it this means we don't really understand this in detail and are just sort of blindly doing what the man pages imply? What does kubernetes do here? Is there discussion in some upstream PR?

plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -76,6 +79,30 @@ add_ovs_flows() {
...
+        egress_burst=${mtu}
+        if [ "${egress_burst}" -lt "$((${egress_bw}/10))" ]; then
+            egress_burst=$((${egress_bw}/10))
+        fi
+        # OVS ingress_policing_burst specified in Kb
+        egress_burst_kb=$((${egress_burst}/1000))
```
danwinship: I think this is right, but it would be clearer (and match the comment better) if you computed `egress_burst_kb` from `mtu` first and then compared it to `egress_kbps` rather than `egress_bw`.

plugins/osdn/ovs/bin/openshift-sdn-ovs
```diff
@@ -76,6 +79,30 @@ add_ovs_flows() {
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=0, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
 	ovs-ofctl -O OpenFlow13 add-flow br0 "table=7, priority=100, reg0=${tenant_id}, ip, nw_dst=${ipaddr}, actions=output:${ovs_port}"
     fi
+
+    # Pod ingress == OVS vSwitch egress
+    # linux-htb used here since that's the Kubernetes default traffic shaper too
+    if [ -n "${ingress_bw}" ]; then
+        qos=$(ovs-vsctl create qos type=linux-htb other-config:max-rate=${ingress_bw})
+        ovs-vsctl set port ${veth_host} qos=${qos}
```
danwinship: what happened to the queue stuff? not needed?

dcbw: Yeah, that was left-over from some earlier experiments and I forgot the queues aren't actually needed unless you want to restrict traffic in a specific pattern.  Just setting a QoS on a specific interface with a max_rate is enough for this case.

plugins/osdn/ovs/bin/openshift-sdn-ovs
dcbw: Kubernetes just sends the rate to the 'tc' tool, and it does some magic and figures things out.  Turns out OVS does some defaults magic and sets the burst to something sane too.  So I don't think we actually need to bother with the burst rate.  See https://github.com/osrg/openvswitch/blob/master/lib/netdev-linux.c#L2612 for what OVS is doing.

dcbw:
@danwinship PTAL

danwinship:
LGTM. anyone else?

PR #267, opened 2016-03-05
Validate cluster/service network only when there is a config change

pravisankar:


pravisankar:
@openshift/networking PTAL

danwinship:
LGTM. This is just an efficiency thing, right? There's nothing actually buggy about the existing code?

dcbw:
LGTM to me too; same question as danw though....

pravisankar:
@danwinship @dcbw 
No issues with the existing code. With this change, validateNetworkConfig() is only called in case of cluster network change and is skipped during start/restart of openshift-master service.

PR #268, opened 2016-03-10
Implement better OVS flow rule versioning

dcbw:
Add a rule stamp and check that stamp on startup to see if we have
to recreate the OVS bridge and flow rules.  Should fix ambiguity
in the existing rule checking.

### DIFF
--- a/plugins/osdn/ovs/controller.go
+++ b/plugins/osdn/ovs/controller.go
@@ -19,6 +19,11 @@ import (
 )
 
 const (
+	// rule versioning; increment each time flow rules change
+	VERSION        = 1
+	VERSION_TABLE  = "table=253"
+	VERSION_ACTION = "actions=note:"
+
 	BR       = "br0"
 	LBR      = "lbr0"
 	TUN      = "tun0"
@@ -33,6 +38,18 @@ type FlowController struct {
 
 func NewFlowController(multitenant bool) *FlowController {
 	return &FlowController{multitenant}
+}
+
+func getPluginVersion(multitenant bool) []string {
+	if VERSION > 254 {
+		panic("Version too large!")
+	}
+	version := fmt.Sprintf("%02X", VERSION)
+	if multitenant {
+		return []string{"01", version}
+	}
+	// single-tenant
+	return []string{"00", version}
 }
 
 func alreadySetUp(multitenant bool, localSubnetGatewayCIDR string) bool {
@@ -63,11 +80,20 @@ func alreadySetUp(multitenant bool, localSubnetGatewayCIDR string) bool {
 	}
 	found = false
 	for _, flow := range flows {
-		if !strings.Contains(flow, "table=3") {
+		if !strings.Contains(flow, VERSION_TABLE) {
 			continue
 		}
-		if (multitenant && strings.Contains(flow, "NXM_NX_TUN_ID")) ||
-			(!multitenant && strings.Contains(flow, "goto_table:9")) {
+		idx := strings.Index(flow, VERSION_ACTION)
+		if idx < 0 {
+			continue
+		}
+
+		// OVS note action format hex bytes separated by '.'; first
+		// byte is plugin type (multi-tenant/single-tenant) and second
+		// byte is flow rule version
+		expected := getPluginVersion(multitenant)
+		existing := strings.Split(flow[idx+len(VERSION_ACTION):], ".")
+		if len(existing) >= 2 && existing[0] == expected[0] && existing[1] == expected[1] {
 			found = true
 			break
 		}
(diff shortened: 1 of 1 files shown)


dcbw:
@danwinship @pravisankar @knobunc

plugins/osdn/ovs/controller.go
```diff
@@ -63,11 +69,16 @@ func alreadySetUp(multitenant bool, localSubnetGatewayCIDR string) bool {
...
-		if !strings.Contains(flow, "table=3") {
+		if !strings.Contains(flow, STAMP_TABLE) {
+			continue
+		}
+		idx := strings.Index(flow, STAMP_ACTION)
+		if idx < 0 {
 			continue
 		}
-		if (multitenant && strings.Contains(flow, "NXM_NX_TUN_ID")) ||
-			(!multitenant && strings.Contains(flow, "goto_table:9")) {
```
danwinship: So, this is supposed to be checking if you switched from single-tenant to multi-tenant or vice versa. Except it doesn't actually work with the current version of the flows. Sigh.

Encode that into the note as well? `0.${STAMP}` for single-tenant rules, `1.${STAMP}` for multi-tenant?

dcbw: Also supposed to check whether you updated openshift and restarted it, and we changed the flow rules in between.  I missed the plugin switch case, so I'll fix that up.

dcbw:
@danwinship PTAL

danwinship:
LGTM

plugins/osdn/ovs/controller.go
```diff
@@ -187,6 +213,10 @@ func (c *FlowController) Setup(localSubnetCIDR, clusterNetworkCIDR, servicesNetw
 	otx.AddPort(TUN, 2, "type=internal")
 	otx.AddPort(VOVSBR, 3)
 
+	// Table 253: rule version stamp; note action is hex bytes separated by '.'
+	pluginStamp := getPluginStamp(c.multitenant)
+	otx.AddFlow("%s, %s%s.%s", STAMP_TABLE, STAMP_ACTION, pluginStamp[0], pluginStamp[1])
+
```
pravisankar: Can we add this flow rule at the end of the Setup()?
For some reason, tun0/docker0 transaction or net sysctl fails (these are executed after br0 transaction), we won't be able to fix setup by restarting the openshift service as alreadySetup() will return true.

danwinship: good idea. also, if this is going to get revised again, maybe change "STAMP" to "VERSION"?

dcbw:
@danwinship @pravisankar PTAL

plugins/osdn/ovs/controller.go
```diff
@@ -294,6 +320,10 @@ func (c *FlowController) Setup(localSubnetCIDR, clusterNetworkCIDR, servicesNetw
 		return false, fmt.Errorf("Could not enable IPv4 forwarding on %s: %s", TUN, err)
 	}
 
+	// Table 253: rule version; note action is hex bytes separated by '.'
+	pluginVersion := getPluginVersion(c.multitenant)
+	otx.AddFlow("%s, %s%s.%s", VERSION_TABLE, VERSION_ACTION, pluginVersion[0], pluginVersion[1])
+
```
pravisankar: Do we want to check if the addFlow succeeded here?
err = otx.EndTransaction()
if err != nil {
	return false, err
}

danwinship: I was going to say that it's unlikely that this would fail if everything before it succeeded, and even if it does, everything would still work fine, we'd just unnecessarily recreate everything next time...

But there is one way this could fail even if everything else succeeded, which is if we stupidly made a typo in this line and caused an ovs syntax error. But the "everything would still work fine" argument still holds. So maybe check the error and just log it if we get one?

Otherwise LGTM

dcbw: Yeah, I did that last week and was testing it but apparently forgot to push it. Will do now.

dcbw:
@pravisankar @danwinship PTAL

pravisankar:
LGTM

bmeng:
@dcbw What does the 2nd byte of the actions=note: do? Was it going to log the history of the openflow rule changes.
But during my testing, I found the value of it will always be "01". It will be overwrote each time when SDN initialized.
The 1st byte will be changed when I changed the plugin type every time. 
Thought the 2nd byte should increase each time re-setup SDN if it is doing some historical work.

danwinship:
@bmeng the second byte indicates the version of the code that was used to create those rules. When you restart OpenShift, it checks if the single-vs-multi-tenant byte and the version byte are both what it was expecting. If they are, then it doesn't need to modify the rules, because they're already correct. But if either of them is wrong, it deletes all the existing rules and creates new ones. (So, the next time we change the rules in `controller.go`, we'll change `VERSION` to "02", and then if you update the binary on your already-deployed system and restart, it will see that your rules there were made with an old version, and so it will recreate them.)

knobunc:
LGTM

knobunc:
BTW, should we use the mask feature of the cookie to make it easy to find all of our rules?  e.g. pick something like 0xbb000000/0xff000000 so that we can easily and cleanly nuke all of our flow rules with:
`ovs-ofctl -OOpenFlow13 dump-flows br0 cookie=0xbb000000/0xff000000`

Or perhaps, to be good ovs doobies, make that cookie value settable in the config?  And provide a utility that reads the config and cleans out the flows?  And if we were feeling clever, we could look in table 253 for the new rule this adds to identify us, and make sure that it is not present with another cookie prefix (implying that a cookie had been changed in the conf).

danwinship:
All of the rules (on br0) are our rules.

knobunc:
Ok, so the ones without the cookie of 0x0 are auto-learned?

danwinship:
There are no auto-learned rules in the current ruleset.

Originally our rules used cookies, but most of them don't any more, because:

    ovs-ofctl -O OpenFlow13 del-flows br0 "ip,nw_dst=${ipaddr}"

is simpler than doing

    ovs-ofctl -O OpenFlow13 del-flows br0 "cookie=0x${cookie}/0xffffffff"

because you don't have to encode the IP address into hex first.

For historical reasons ("no one ever bothered to change it"), the per-node subnet routing rules still use a cookie though.

knobunc:
Got it.  Thanks!

PR #270, opened 2016-03-22
plugin refactoring/cleanup

danwinship:
The individual commit messages mostly explain the changes here...

The change to get rid of the wrapper types was inspired by the egress router work: I wanted to be able to check the pod's SecurityContext from plugin.go, which would have required copying that info from kapi.Pod to osdnapi.Pod, much as dcbw had just done for annotations... but that seemed silly when we could just use the kapi.Pod directly instead.

@openshift/networking PTAL

### DIFF
--- a/plugins/osdn/common.go
+++ b/plugins/osdn/common.go
@@ -22,6 +22,15 @@ import (
 type PluginHooks interface {
 	PluginStartMaster(clusterNetwork *net.IPNet, hostSubnetLength uint) error
 	PluginStartNode(mtu uint) error
+
+	SetupSDN(localSubnetCIDR, clusterNetworkCIDR, serviceNetworkCIDR string, mtu uint) (bool, error)
+
+	AddOFRules(nodeIP, nodeSubnetCIDR, localIP string) error
+	DelOFRules(nodeIP, localIP string) error
+
+	AddServiceOFRules(netID uint, IP string, protocol api.ServiceProtocol, port uint) error
+	DelServiceOFRules(netID uint, IP string, protocol api.ServiceProtocol, port uint) error
+
 	UpdatePod(namespace string, name string, id kubetypes.DockerID) error
 }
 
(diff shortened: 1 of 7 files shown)


dcbw:
LGTM

plugins/osdn/common.go
```diff
@@ -26,11 +26,11 @@ type PluginHooks interface {
...
 
-	AddOFRules(nodeIP, nodeSubnetCIDR, localIP string) error
-	DelOFRules(nodeIP, localIP string) error
+	HostSubnetAdded(subnet *osapi.HostSubnet)
+	HostSubnetDeleted(subnet *osapi.HostSubnet)
 
-	AddServiceOFRules(netID uint, IP string, protocol kapi.Protocol, port int) error
-	DelServiceOFRules(netID uint, IP string, protocol kapi.Protocol, port int) error
+	ServiceAdded(service *kapi.Service, netID uint)
+	ServiceDeleted(service *kapi.Service, netID uint)
```
pravisankar: AddHostSubnet, DeleteHostSubnet, AddService, DeleteService names will be better. These will denote intent instead of completed action names and also it will be inline with SetupSDN and UpdatePod (action+object).

pravisankar:
Minor nitpick, LGTM

plugins/osdn/common.go
danwinship: The intent of the function isn't "add a HostSubnet" though. It's called after the HostSubnet has already been added, so the intent-based name would be "UpdateStateForAddedHostSubnet" or something.

(The old names were intent-based, but I was trying to switch to names that didn't presuppose the use of OpenFlow rules...)

danwinship:
@pravisankar redid the last commit to do less renaming so the names are still intent based. They're now "AddHostSubnetRules"/"AddServiceRules" rather than "AddOFRules"/"AddServiceOFRules". LGTY?

pravisankar:
I like the updated method names.
Also update the comments where these names are referred:
https://github.com/danwinship/openshift-sdn/blob/refactor/plugins/osdn/ovs/controller.go#L227
https://github.com/danwinship/openshift-sdn/blob/refactor/plugins/osdn/ovs/controller.go#L241
https://github.com/danwinship/openshift-sdn/blob/refactor/plugins/osdn/ovs/controller.go#L265

plugins/osdn/subnets.go
```diff
@@ -123,21 +128,23 @@ func (oc *OvsController) SubnetStartNode(mtu uint) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	subnets := result.([]api.Subnet)
+	subnets := result.([]osapi.HostSubnet)
 	for _, s := range subnets {
-		oc.flowController.AddOFRules(s.NodeIP, s.SubnetCIDR, oc.localIP)
+		if s.HostIP != oc.localIP {
+			oc.pluginHooks.AddHostSubnetRules(&s)
```
pravisankar: Do we want to return error from AddHostSubnetRules()?  watchSubnets() can log the error and SubnetStartNode() can propagate the error to the higher layers?

danwinship: If OVS is totally messed up on the machine, then we'd get errors from the OVS methods during initial setup, and those errors do get checked.

If we fail here because of a bug (syntax error in the rule, bug in pkg/ovs, etc) then that should make networking not work correctly, which should then get caught by the unit tests and then fixed, so the failure would never happen in the real world.

So basically, errors from these functions are sort of "can't happen". The functions used to return errors which all callers always ignored; it seemed to make more sense to remove the errors than to make the callers check them...

danwinship:
> Also update the comments where these names are referred

oops, fixed

pravisankar:
LGTM

PR #271, opened 2016-03-22
Add annotation for adding macvlan interface to pod (for egress-router)

danwinship:
This builds on the refactoring in #270; only the final commit ("Implement "openshift.io/macvlan" annotation on pods") is new

With this addition, you can run https://github.com/danwinship/egress-router, and this should be considered in that context.

We had talked before about whether implementing egress IPs as a pod made sense, and I eventually decided that it does:
- It provides an obvious way to enable/disable the service.
- It keeps the logic outside of openshift-sdn/origin, so we can deprecate/replace it later if we need to.
- You can point a service at the pod and then other pods refer to it by that and don't need to know the details of where it's running.
- You can use a nodeSelector on the pod to force it to only be deployed to nodes that are capable of using the egress IP.
- If the node running the egress router pod goes down, it can automatically get deployed to another node. (There had been discussion of whether it would make more sense to tie that into the existing "high availability" stuff, but that is about keeping publicly-visible IPs working, and doesn't really match what we're doing here.)
- Isolation is already taken care of
- If the functionality is implemented in certain ways, then everything gets cleaned up automatically when the pod's network namespace is deleted, even if the pod was killed uncleanly.

The one catch is that you end up wanting the pod to have its own network namespace and get assigned a ClusterNetwork IP address by openshift-sdn, but you also need to run a small bit of setup code in the host's network namespace. So that's what this PR does.

Some open questions:
- Is the annotation name appropriate? (We already have some other "openshift.io/" annotations, such as "openshift.io/scc").
- Is the "privileged" requirement appropriate? The pod really only needs CAP_NET_ADMIN (I think; not tested), but none of the preexisting SecurityContextConstraints will let you add capabilities to your container SecurityContext, so if we did it that way, you'd need to create a new SCC before you could create the router. (Or we could add it to the default set I guess.)
- Is this feature the right one to add? Another possibility would be, rather than adding code to openshift-sdn-ovs, to allow the pod to provide its own additional startup script to run (sort of like the existing PostStart hook, except running outside the container). That might also be useful for other features, but also seems pretty dangerous... (I'm not sure if the macvlan feature is likely to be useful for anything other than the egress router...)

@openshift/networking PTAL

### DIFF
--- a/plugins/osdn/ovs/bin/openshift-sdn-ovs
+++ b/plugins/osdn/ovs/bin/openshift-sdn-ovs
@@ -8,6 +8,7 @@ net_container=$2
 tenant_id=$3
 ingress_bw=$4
 egress_bw=$5
+macvlan=$6
 
 lockwrap() {
     (
@@ -115,6 +116,16 @@ ensure_subnet_route() {
     add_subnet_route
 }
 
+add_macvlan() {
+    default_dev=$(ip route show | sed -ne 's/^default .* dev \([^ ]*\) .*/\1/p')
+    if [ -z "$default_dev" ]; then
+	echo "Could not find default network interface"
+	exit 1
+    fi
+    ip link add link $default_dev name macvlan0 type macvlan mode private
+    ip link set macvlan0 netns $pid
+}
+
 run() {
     get_ipaddr_pid_veth
     source /run/openshift-sdn/config.env
@@ -124,6 +135,9 @@ run() {
 	    add_ovs_port
 	    add_ovs_flows
 	    add_subnet_route
+	    if [ "$macvlan" = true ]; then
+		add_macvlan
+	    fi
 	    ;;
 
 	update)
--- a/plugins/osdn/ovs/plugin.go
+++ b/plugins/osdn/ovs/plugin.go
@@ -162,6 +162,19 @@ func extractBandwidthResources(pod *kapi.Pod) (ingress, egress int64, err error)
 	return ingress, egress, nil
 }
 
+func wantsMacvlan(pod *kapi.Pod) (bool, error) {
+	val, found := pod.Annotations["pod.network.openshift.io/assign-macvlan"]
+	if !found || val != "true" {
+		return false, nil
+	}
+	for _, container := range pod.Spec.Containers {
+		if container.SecurityContext.Privileged != nil && *container.SecurityContext.Privileged {
+			return true, nil
+		}
+	}
+	return false, fmt.Errorf("Pod has 'pod.network.openshift.io/assign-macvlan' annotation but is not privileged")
+}
+
 func (plugin *ovsPlugin) SetUpPod(namespace string, name string, id kubeletTypes.DockerID) error {
 	err := plugin.WaitForPodNetworkReady()
 	if err != nil {
@@ -192,7 +205,12 @@ func (plugin *ovsPlugin) SetUpPod(namespace string, name string, id kubeletTypes
 		return err
 	}
 
-	out, err := utilexec.New().Command(plugin.getExecutable(), setUpCmd, string(id), vnidstr, ingressStr, egressStr).CombinedOutput()
+	macvlan, err := wantsMacvlan(pod)
+	if err != nil {
+		return err
+	}
+
+	out, err := utilexec.New().Command(plugin.getExecutable(), setUpCmd, string(id), vnidstr, ingressStr, egressStr, fmt.Sprintf("%t", macvlan)).CombinedOutput()
 	glog.V(5).Infof("SetUpPod network plugin output: %s, %v", string(out), err)
 	return err
 }


knobunc:
Nice work.  LGTM.

Regarding the SCC, I need a CAP_NET_ADMIN one too for the SYN eater, I need to hammer out what @smarterclayton thinks is the best path forward there.

smarterclayton:
For the annotation (which is an API) would suggest you add more bits: `network.openshift.io/pod-macvlan` (or something a bit broader).

Is CAP_NET_ADMIN dangerous enough that we would *always* need it to be privileged, or could we check for both privileged or you are given the cap?  I.e. is there anything besides CAP_NET_ADMIN like SELinux that would prevent the vlan from being set?

danwinship:
> Is CAP_NET_ADMIN dangerous enough that we would always need it to be privileged

In the case of a normal pod, CAP_NET_ADMIN isn't really dangerous at all, because it only gives you power within the pod's network namespace. Our OVS rules only accept outgoing traffic with the correct IP and MAC addresses, so nothing you do in your own network namespace can affect what traffic gets out of the pod.

CAP_NET_ADMIN plus a macvlan is more dangerous: It gives the pod the ability to spoof/sniff traffic on the node's subnet.

So maybe that's an argument for making the annotation require privileged, not CAP_NET_ADMIN; the annotation makes CAP_NET_ADMIN become more dangerous than it used to be (and more dangerous than it seems like it should be).

danwinship:
> For the annotation (which is an API) would suggest you add more bits: `network.openshift.io/pod-macvlan` (or something a bit broader).

You mean, give it a more specific name so it doesn't conflict with other future feature ideas, right? OK, re-pushed with the annotation changed to `openshift-sdn.openshift.io/add-macvlan` ("`openshift-sdn`" rather than "`network`" because we don't necessarily expect other network plugins to implement the same feature [right?] and "`add-macvlan`" rather than "`pod-macvlan`" because "pod" is already implied by the fact that this is an annotation on a pod, and it's about adding a macvlan interface, not replacing the default veth interface, etc)

smarterclayton:
Think of the annotation like an API.  openshift.io is the default API group
- the object or sub object it applies to is the prefix to openshift.io.
The value after the slash corresponds to the field name.

If someone else had a network plugin, roughly comparable to openshift-sdn,
would they want to implement this?  Is this a feature that *only* makes
sense on our particular SDN implementation?  If so, then making it specific
is fine.  If not, then having it be specific is bad.  You want to make sure
that you identify that this is related to pods, so "pod" needs to be in the
name somewhere.

  "pod.network.openshift.io/assign-macvlan": "true"

is probably more generic (we don't use "add" for singletons in our public
APIs usually).  Or

  "pod.sdn.openshift.io/assign-macvlan" (openshift-sdn.openshift.io is
redundant)

danwinship:
> If someone else had a network plugin, roughly comparable to openshift-sdn, would they want to implement this? Is this a feature that *only* makes sense on our particular SDN implementation?

The feature is not specific to openshift-sdn. In fact, even the *implementation* isn't specific to openshift-sdn... any other plug-in would implement it in exactly the same way, so I guess ideally it would happen outside of openshift-sdn. But we don't have any good place to do that right now. (I guess we could have the origin node startup code create a wrapper NetworkPlugin object around the actual network plugin, so that origin could insert its own plugin-independent network setup stuff... But that could happen later.)

Re-pushed with the annotation changed to `pod.network.openshift.io/assign-macvlan`

danwinship:
@smarterclayton also, are you ok with https://github.com/danwinship/egress-router, and if so, what would be the procedure for getting that imported as openshift/egress-router?

smarterclayton:
For egress-router, do you anticipate having tons of code in it?  If not,
would recommend moving it into origin as images/router/egress and then
we'll add it to all the official builds and manage that centrally (since
it's small).

Will take a look at this pull before EoD, didn't have many comments beyond
that so if you have signoff from the others on this I probably won't have
much.

dcbw:
I wonder if, since the macvlan stuff is egress-router specific, we shouldn't rename the annotation to be specific to the egress router?  The annotation could be abused pretty easily, and tying it to the egress router even if in name only might reduce that.

danwinship:
It's true we don't have any other use cases lined up for it, but it's at least theoretically useful for other things...

As for abuse, that's why it requires "privileged"; so that only people who you trust to not abuse it can make use of it.

The corresponding origin branch is [merge]-tagged now so I guess we should get this in and then sync it over to origin?

simongun:
@danwinship, in the [Egress router doc you write](https://github.com/danwinship/egress-router)

> The pod.network.openshift.io/assign-macvlan annotation tells OpenShift to create a macvlan network interface on the primary network interface

Any reason to restrict this to the primary interface? My concern is that on AWS (e.g. m2.2xlarge), we only get 30 IPs per interface. Considering we might have to use 2 IPs per egress service (for failover purposes), this will not be enough.

Could the egress router make use of all interfaces (4 in the case of AWS m2.2xlarge)?

danwinship:
@simongun what is the use case you're considering here where you expect to need so many egress routers?

simongun:
@danwinship In a public cloud environment, each project is very likely to consume 1..n (where n is likely ~3) external services, such as Object Store, key-value stores, databases. We also might want to identify a project (based on it's IP) towards additional services such as API gateways or an HTTP proxy. Assuming that we will have > 5 projects in average per m2.2xlarge, we are already exceeding the limit of 15 IPs (30 div 2, because of the resiliency requirement).

danwinship:
The egress-router feature wasn't really intended for general use. (Note that egress-router pods can only be created by cluster administrators.) In particular, it was designed for use in cases where having a fixed outgoing IP was *mandatory* (ie, imposed on you by someone else), and so OpenShift was unusable without it. If we expected every project to have an egress IP we'd want to do this differently. (But we don't expect that... cloud services are cloudy.)

That said, making the code support macvlans off other interfaces would be pretty simple. You could make the annotation support either "true" or an interface name, for instance. Patches welcome.

danwinship:
also, note that there's another feature in progress for outbound firewalling

PR #272, opened 2016-03-24
Fix qos check in del_ovs_flows()

pravisankar:
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1320430


pravisankar:
@openshift/networking PTAL

danwinship:
how exactly does this bug cause the reported behavior?

dcbw:
It could cause the bug by exiting the script early because qos = "[]" and that QOS doesn't exist, so "ovs-vsctl destroy qos []" returns 1.  This fix LGTM, but we should also do:

ovs-vsctl --if-exists destroy qos ${qos}

to be slightly safer.  The 'clear port' is safe becuase it doesn't care if QoS exists or not.

pravisankar:
@dcbw updated

PR #273, opened 2016-03-24
Improved the debug script to grab the docker unit file.

knobunc:
This changes the debug script to call systemctl to get the complete docker unit file and then finds out the docker run file used, and pulls the contents of that to get the interface config.


pravisankar:
Looks good.

PR #278, opened 2016-04-01
Fix broken VnidStartMaster()

pravisankar:


pravisankar:
@openshift/networking PTAL

pravisankar:
Fixes panic: interface conversion: interface is []api.Namespace, not []string

danwinship:
that code hasn't changed since October... why hasn't this been a problem before?

pravisankar:
Earlier registry.GetNamespaces() returned []string and as part of this refactor https://github.com/openshift/openshift-sdn/pull/270 we started returning []kapi.Namespace and we missed this change during the refactor.

knobunc:
LGTM. @danwinship any complaints?

PR #282, opened 2016-04-04
[DO NOT MERGE] subnet: add locking to fix duplicate subnet allocations (rh #1323275)

dcbw:
The subnet allocator's GetNetwork() function (which allocates and returns
a subnet for a node) can be called from multiple goroutines and thus
needs locking to prevent duplicate subnet allocations.

### DIFF
--- a/pkg/netutils/subnet_allocator.go
+++ b/pkg/netutils/subnet_allocator.go
@@ -3,6 +3,7 @@ package netutils
 import (
 	"fmt"
 	"net"
+	"sync"
 )
 
 type SubnetAllocator struct {
@@ -14,6 +15,7 @@ type SubnetAllocator struct {
 	rightMask  uint32
 	next       uint32
 	allocMap   map[string]bool
+	mux        sync.Mutex
 }
 
 func NewSubnetAllocator(network string, hostBits uint, inUse []string) (*SubnetAllocator, error) {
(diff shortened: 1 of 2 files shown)


danwinship:
How does it get called from multiple goroutines?

Given that the bz bug mentions a HA setup, I assumed the problem was that we were running the code on multiple masters at once... (though I had thought someone convinced me at one point that that couldn't actually happen, for some reason that I don't remember).

pravisankar:
I do not understand why we need this change?
watchNodes() waits for node events and these events are processed *serially* and may call addNode()/deleteNode() which in turn will use subnet allocator to allocate/release subnetIP. Synchronization is implicitly provided by the event queue.

Is the HA setup right? RunSDNController() as part of openshift startControllers should be run on only one of the HA masters otherwise we will have race conditions, even this mutex lock won't help as this will provide synchronization within a single master but not across masters.

dcbw:
> Is the HA setup right? RunSDNController() as part of openshift startControllers should be run on only one of the HA masters otherwise we will have race conditions, even this mutex lock won't help as this will provide synchronization within a single master but not across masters.

You are correct; I missed the HA angle here and the reporter is definitely running HA.  Doesn't the requirement of a single SDN master kinda subvert HA though?  Do we need to modify openshift-ansible to configure only the first master with the SDN plugin?

PR #283, opened 2016-04-04
Added 'docker ps -a' to the node debug items

knobunc:
It would have been handy to see the docker state vs. the openshift state for containers for a bug I was working on.  This addresses that deficiency.


knobunc:
@dcbw @danwinship PTAL

PR #285, opened 2016-04-07
Fixes the ip address to namespace cache when pods reuse addresses

knobunc:
There's a cache that tracks which namespace IP addresses belong to.  However, the cache logic was not considering that Failed or Complete pods (i.e. not running) could share an IP address with a Running pod.  so the cache code needed to be changed to only add items to the cache when a pod was Running or Pending.  We also track the pod UID to make sure that when we delete, we are deleting the entry for the pod that was running.  So that if a stopped pod is edited, that entry doesn't end up in the cache.

This is reproducible by setting the networkConfig hostSubnetLength to 3 in the master and node config on your cluster.  Then create two namespaces, ns1 and ns2, a service in ns2 that has a selector that matches both pods then create a pod that fails in one (have it run "sleep 5; false", and set the restartPolicy to Never.  Once it has failed, create a pod that runs correctly in the other (hello-openshift is fine).  Then curling the service URL should work.  Remove the failed pod and the curl will then fail.

The problem is that the cache sees the failed pod removal and removes the ip from the ip->namespace cache.  But it should not have been removed because the running pod put the entry in.

For bug:
  https://bugzilla.redhat.com/show_bug.cgi?id=132294

Also-Authored-By: sross@redhat.com

### DIFF
--- a/plugins/osdn/registry.go
+++ b/plugins/osdn/registry.go
@@ -31,7 +31,7 @@ import (
 type Registry struct {
 	oClient          osclient.Interface
 	kClient          kclient.Interface
-	namespaceOfPodIP map[string]string
+	podsByIP         map[string]*kapi.Pod
 	serviceNetwork   *net.IPNet
 	clusterNetwork   *net.IPNet
 	hostSubnetLength int
@@ -75,9 +75,9 @@ type ServiceEvent struct {
 
 func NewRegistry(osClient *osclient.Client, kClient *kclient.Client) *Registry {
 	return &Registry{
-		oClient:          osClient,
-		kClient:          kClient,
-		namespaceOfPodIP: make(map[string]string),
+		oClient:  osClient,
+		kClient:  kClient,
+		podsByIP: make(map[string]*kapi.Pod),
 	}
 }
 
@@ -137,7 +137,7 @@ func (registry *Registry) GetPods() ([]kapi.Pod, string, error) {
 
 	for _, pod := range podList.Items {
 		if pod.Status.PodIP != "" {
-			registry.namespaceOfPodIP[pod.Status.PodIP] = pod.ObjectMeta.Namespace
+			registry.trackPod(&pod)
 		}
 	}
 	return podList.Items, podList.ListMeta.ResourceVersion, nil
@@ -156,9 +156,9 @@ func (registry *Registry) WatchPods(ready chan<- bool, start <-chan string, stop
 
 		switch eventType {
 		case watch.Added, watch.Modified:
-			registry.namespaceOfPodIP[pod.Status.PodIP] = pod.ObjectMeta.Namespace
+			registry.trackPod(pod)
 		case watch.Deleted:
-			delete(registry.namespaceOfPodIP, pod.Status.PodIP)
+			registry.unTrackPod(pod)
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


knobunc:
@sross PTAL

DirectXMan12:
@knobunc :+1: thanks!

knobunc:
@danwinship @pravisankar PTAL

danwinship:
> There's a cache that tracks which namespace IP addresses belong to. However, the cache logic was not considering that Failed or Complete pods (i.e. not running) could share an IP address with a Running pod.

I still feel like *this* is the bug, and we ought to at least file a bug against kubernetes... (even while working around the problem here)

plugins/osdn/registry.go
```diff
@@ -666,3 +671,39 @@ EndpointLoop:
 
 	registry.baseEndpointsHandler.OnEndpointsUpdate(filteredEndpoints)
 }
+
+func (registry *Registry) trackPod(pod *kapi.Pod) {
+	if pod.Status.PodIP != "" {
```
danwinship: could simplify this by just

    if pod.Status.PodIP == "" {
        return
    }

plugins/osdn/registry.go
```diff
@@ -666,3 +671,39 @@ EndpointLoop:
...
+			// we need to add it to our IP -> namespace tracker.  There _should_ be no
+			// other entries for the IP if we caught all of the right messages, so warn
+			// if we see one, but clobber it anyway since the IPAM
+			// should ensure that each IP is uniquely assigned to a pod (when running)
+			if ok && info.uid != pod.UID {
+				log.Warningf("IP '%s' was marked as used by namespace '%s' (pod '%s')... updating to namespace '%s' (pod '%s')",
+					pod.Status.PodIP, info.namespace, info.uid, pod.ObjectMeta.Namespace, pod.UID)
+			}
+
+			registry.namespaceOfPodIP[pod.Status.PodIP] = podInfo{namespace: pod.ObjectMeta.Namespace, uid: pod.UID}
```
danwinship: maybe just store the *kapi.Pod itself for simplicity? Also, "namespaceOfPodIP" is no longer an accurate description of the data...

danwinship:
basically looks good

plugins/osdn/registry.go
pravisankar: Yes, change 'namespaceOfPodIP' to 'podInfoMap' or some other better name

pravisankar:
LGTM after danwinship's comments are addressed

knobunc:
@danwinship @pravisankar @DirectXMan12 PTAL

knobunc:
@danwinship consensus on slack #kubernetes-dev (@ncdc et al) was that multiple pods can have the same podIP as long as only one is running at a time.

pravisankar:
lgtm

smarterclayton:
Was this an existing issue that no one encountered?  Or did we introduce it with other changes?  If so, what release was this introduced?

danwinship:
It doesn't seem to be caused by any change in openshift-sdn, but it also seems like people would have noticed this bug sooner if it had always been around. So either it's a regression caused by changes outside of openshift-sdn, or else it's just a sign that not many people were using multitenant before.

PR #286, opened 2016-04-07
Add more logging, to help debug problems

danwinship:
@openshift/networking PTAL

### DIFF
--- a/plugins/osdn/ovs/plugin.go
+++ b/plugins/osdn/ovs/plugin.go
@@ -34,7 +34,7 @@ func MultiTenantPluginName() string {
 func CreatePlugin(registry *osdn.Registry, multitenant bool, hostname string, selfIP string) (api.OsdnPlugin, api.FilteringEndpointsConfigHandler, error) {
 	plugin := &ovsPlugin{multitenant: multitenant}
 
-	err := plugin.BaseInit(registry, plugin, hostname, selfIP)
+	err := plugin.BaseInit(registry, plugin, multitenant, hostname, selfIP)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/plugins/osdn/subnets.go
+++ b/plugins/osdn/subnets.go
@@ -59,7 +59,7 @@ func (oc *OsdnController) SubnetStartMaster(clusterNetwork *net.IPNet, hostSubne
 		}
 		_, err = oc.Registry.GetSubnet(node.Name)
 		if err == nil {
-			// subnet already exists, continue
+			log.V(5).Infof("HostSubnet for node %q already exists", node.Name)
 			continue
 		}
 		err = oc.addNode(node.Name, nodeIP)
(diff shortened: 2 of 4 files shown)


plugins/osdn/ovs/plugin.go
```diff
@@ -34,7 +34,7 @@ func MultiTenantPluginName() string {
 func CreatePlugin(registry *osdn.Registry, multitenant bool, hostname string, selfIP string) (api.OsdnPlugin, api.FilteringEndpointsConfigHandler, error) {
 	plugin := &ovsPlugin{multitenant: multitenant}
 
-	err := plugin.BaseInit(registry, plugin, hostname, selfIP)
+	err := plugin.BaseInit(registry, plugin, multitenant, hostname, selfIP)
 	if err != nil {
 		return nil, nil, err
 	}
```
pravisankar: We could log multi-tent plugin info here as BaseInit doesn't need multitenant other than logging.

plugins/osdn/subnets.go
```diff
@@ -71,37 +71,40 @@ func (oc *OsdnController) SubnetStartMaster(clusterNetwork *net.IPNet, hostSubne
...
+		return fmt.Errorf("Error allocating network for node %s: %v", nodeName, err)
+	}
 	err = oc.Registry.CreateSubnet(nodeName, nodeIP, sn.String())
 	if err != nil {
-		log.Errorf("Error writing subnet to etcd for node %s: %v", nodeName, sn)
-		return err
+		return fmt.Errorf("Error writing subnet %v to etcd for node %s: %v", sn, nodeName, err)
 	}
+
+	log.V(5).Infof("Created HostSubnet %v for node %s (%s)", sn, nodeName, nodeIP)
```
pravisankar: We don't expect a ton of these anyway, can we make it log.Infof?

plugins/osdn/subnets.go
```diff
@@ -71,37 +71,40 @@ func (oc *OsdnController) SubnetStartMaster(clusterNetwork *net.IPNet, hostSubne
...
+		return fmt.Errorf("Error parsing subnet %q for node %q for deletion: %v", sub.Subnet, nodeName, err)
 	}
 	oc.subnetAllocator.ReleaseNetwork(ipnet)
-	return oc.Registry.DeleteSubnet(nodeName)
+	err = oc.Registry.DeleteSubnet(nodeName)
+	if err != nil {
+		return fmt.Errorf("Error deleting subnet %v for node %q: %v", sub, nodeName, err)
+	}
+
+	log.V(5).Infof("Deleted HostSubnet %v for node %s", sub, nodeName)
```
pravisankar: same comment as before, log.Infof?

plugins/osdn/ovs/plugin.go
danwinship: Yeah, it's dumb, but I didn't want to log two separate lines... (We can't log hostname and selfIP from here because in some cases they get computed by BaseInit(), and the computed IP isn't stored in a public struct member, so we can't access it after BaseInit() returns either. So we have to log that from there.) But I could change that.

pravisankar: I'm fine with this change.

danwinship:
done, and likewise with the vnid allocation/revocation

pravisankar:
lgtm

PR #287, opened 2016-04-07
Added the master and node config files to the debug tar

knobunc:
The configuration was not included in the debug tars and are often helpful when debugging issues.  This changes adds them.

### DIFF
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -52,6 +52,7 @@ log_system () {
     echo_and_eval  cat /etc/resolv.conf                               &> $logpath/resolv.conf
     echo_and_eval  lsmod                                              &> $logpath/modules
     echo_and_eval  sysctl -a                                          &> $logpath/sysctl
+    echo_and_eval  cat "$config"                                      &> $logpath/master-config.yaml
 
     echo_and_eval  oc version                                         &> $logpath/version
     echo                                                             &>> $logpath/version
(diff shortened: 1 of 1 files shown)


hack/debug.sh
```diff
@@ -61,14 +62,25 @@ log_system () {
 }
 
 do_master () {
+    # Find the config file
+    config=$(ps wwaux | grep -v grep | sed -ne 's/.*openshift.*--\(master-\)\?config=\([^ ]*\.yaml\).*/\2/p')
```
danwinship: Our normal configuration doesn't use a config file for the master, it just passes stuff on the command line. Also, if the master is also a node, then this expression will match the node process and end up grabbing its config file.

hack/debug.sh
```diff
@@ -61,14 +62,25 @@ log_system () {
...
+    if [ -z "$config" ]; then
+      config=$(systemctl show -p ExecStart $aos_master_service | sed -ne 's/.*--\(master-\)\?config=\([^ ]*\).*/\1/p')
+    fi
+    if [ -z "$config" ]; then
+      die "Could not find master-config.yaml from 'ps' or 'systemctl show'"
```
danwinship: If the master invocation doesn't use --config, then just output the "ps" output instead. (Or the systemctl show output I guess for consistency, although it's already present in the debug data anyway...)

hack/debug.sh
```diff
@@ -61,14 +62,25 @@ log_system () {
...
+    fi
+
+    # Pull all of the nodes from the system
     if ! nodes=$(oc get nodes --template '{{range .items}}{{.spec.externalID}} {{end}}'); then
-	if [ -z "$KUBECONFIG" -o ! -f "$KUBECONFIG" ]; then
+      if [ -z "$KUBECONFIG" -o ! -f "$KUBECONFIG" ]; then
```
danwinship: gratuitous (incorrect) whitespace changes

hack/debug.sh
```diff
@@ -249,16 +261,19 @@ do_pod_service_connectivity_check () {
 }
 
 do_node () {
-    config=$(ps wwaux | grep -v grep | sed -ne 's/.*openshift.*--config=\([^ ]*\.yaml\).*/\1/p')
+    # Find the config file
+    config=$(ps wwaux | grep -v grep | sed -ne 's/.*openshift.*--\(node-\)\?config=\([^ ]*\.yaml\).*/\1/p')
```
danwinship: of course, this (both before and after) has the same problem as the master code above, in reverse: if run on the master, and the master process was started with --config, then this might pick up the master config rather than the node config...

hack/debug.sh
```diff
@@ -52,6 +52,7 @@ log_system () {
     echo_and_eval  cat /etc/resolv.conf                               &> $logpath/resolv.conf
     echo_and_eval  lsmod                                              &> $logpath/modules
     echo_and_eval  sysctl -a                                          &> $logpath/sysctl
+    echo_and_eval  cat "$config"                                      &> $logpath/master-config.yaml
```
pravisankar: This puts the assumption that log_system() is always called from do_master() or do_node()?  and in case of node, master-config.yaml is confusing, it should be node-config.yaml?
Why don't we capture and log this info in do_master() and do_node() itself?

PR #288, opened 2016-04-08
debug.sh fixes for vxlan anti-spoofing

danwinship:
Our OpenFlow rules now try to prevent vxlan spoofing, but the test packet traces in debug.sh didn't get updated for that, so they're now seen as being spoofed and get dropped when they shouldn't

(also another minor fix for dind cluster support)

@openshift/networking


danwinship:
meh, don't merge yet, there's another fix coming

dcbw:
LGTM

PR #292, opened 2016-04-12
Add a mutex to registry.podsByIP

### LINKED ISSUE #291: fatal error with multi-tenant plugin
See this [fpaste](https://paste.fedoraproject.org/354655/46046660/) for the backtrace.

The first few lines are inlined below.

```
fatal error: concurrent map read and map write
goroutine 470 [running]:
runtime.throw(0x38e04e0, 0x21)
/usr/local/go/src/runtime/panic.go:530 +0x90 fp=0xc821789920 sp=0xc821789908
runtime.mapaccess2_faststr(0x2823780, 0xc8207bd4a0, 0xc8211fa9f8, 0x8, 0x1, 0x1)
/usr/local/go/src/runtime/hashmap_fast.go:307 +0x5b fp=0xc821789980 sp=0xc821789920
github.com/openshift/openshift-sdn/plugins/osdn.(*Registry).OnEndpointsUpdate(0xc820652cd0, 0xc821b92000, 0x19, 0x25)
/go/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/openshift/openshift-sdn/plugins/osdn/registry.go:692 +0x74a fp=0xc821789e40 sp=0xc821789980
k8s.io/kubernetes/pkg/proxy/config.(*EndpointsConfig).RegisterHandler.func1(0x27c4cc0, 0xc820b2b6c0)
/go/src/github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/pkg/proxy/config/config.go:100 +0xdc fp=0xc821789ea0 sp=0xc821789e40
k8s.io/kubernetes/pkg/util/config.ListenerFunc.OnUpdate(0xc82090fda0, 0x27c4cc0, 0xc820b2b6c0)
/go/src/github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/pkg/util/config/config.go:110 +0x30 fp=0xc821789eb8 sp=0xc821789ea0
k8s.io/kubernetes/pkg/util/config.(*Broadcaster).Notify(0xc8206bb5c0, 0x27c4cc0, 0xc820b2b6c0)
/go/src/github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/pkg/util/config/config.go:138 +0xf8 fp=0xc821789f40 sp=0xc821789eb8
k8s.io/kubernetes/pkg/proxy/config.watchForUpdates(0xc8206bb5c0, 0x7f53a4922cc8, 0xc8206bb470, 0xc8209c4d80)
/go/src/github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/pkg/proxy/config/config.go:297 +0x76 fp=0xc821789f70 sp=0xc821789f40
runtime.goexit()
/usr/local/go/src/runtime/asm_amd64.s:1998 +0x1 fp=0xc821789f78 sp=0xc821789f70
created by k8s.io/kubernetes/pkg/proxy/config.NewEndpointsConfig
/go/src/github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/pkg/proxy/config/config.go:93 +0x248
```

danwinship:
The pod-monitoring and endpoint-filtering threads might both run at the same time so we need a mutex here. Fixes #291.

As for the actual race condition, if a pod which is an endpoint is added, and OnEndpointsUpdate() ends up running before trackPod(), then the pod won't be allowed as an endpoint and will be ignored until the
next time OnEndpointsUpdate() runs, when it will get picked up. So isolation is guaranteed but correct functioning of the service is not...

But can that actually happen? It seems unlikely unless there's really unfair thread scheduling... (The more common case is that trackPod/unTrackPod and OnEndpointsUpdates are running at the same time for unrelated reasons.) Anyway, we don't like this code and want it to go away and be replaced by something else anyway...

@openshift/networking PTAL

### DIFF
--- a/plugins/osdn/registry.go
+++ b/plugins/osdn/registry.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"sync"
 
 	log "github.com/golang/glog"
 
@@ -28,6 +29,7 @@ type Registry struct {
 	oClient          *osclient.Client
 	kClient          *kclient.Client
 	podsByIP         map[string]*kapi.Pod
+	podTrackingLock  sync.Mutex
 	serviceNetwork   *net.IPNet
 	clusterNetwork   *net.IPNet
 	hostSubnetLength int
@@ -472,7 +474,7 @@ EndpointLoop:
 					continue EndpointLoop
 				}
 				if clusterNetwork.Contains(IP) {
-					podInfo, ok := registry.podsByIP[addr.IP]
+					podInfo, ok := registry.getTrackedPod(addr.IP)
 					if !ok {
 						log.Warningf("Service '%s' in namespace '%s' has an Endpoint pointing to non-existent pod (%s)", ep.ObjectMeta.Name, ns, addr.IP)
 						continue EndpointLoop
@@ -490,11 +492,21 @@ EndpointLoop:
 	registry.baseEndpointsHandler.OnEndpointsUpdate(filteredEndpoints)
 }
 
+func (registry *Registry) getTrackedPod(ip string) (*kapi.Pod, bool) {
+	registry.podTrackingLock.Lock()
+	defer registry.podTrackingLock.Unlock()
+
+	pod, ok := registry.podsByIP[ip]
+	return pod, ok
+}
+
 func (registry *Registry) trackPod(pod *kapi.Pod) {
 	if pod.Status.PodIP == "" {
 		return
 	}
 
+	registry.podTrackingLock.Lock()
+	defer registry.podTrackingLock.Unlock()
 	podInfo, ok := registry.podsByIP[pod.Status.PodIP]
 
 	if pod.Status.Phase == kapi.PodPending || pod.Status.Phase == kapi.PodRunning {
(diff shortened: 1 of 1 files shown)


knobunc:
LGTM

plugins/osdn/registry.go
```diff
@@ -145,12 +149,14 @@ func (registry *Registry) WatchPods() error {
...
 		pod := obj.(*kapi.Pod)
 
+		registry.podTrackingLock.Lock()
 		switch eventType {
 		case watch.Added, watch.Modified:
 			registry.trackPod(pod)
 		case watch.Deleted:
 			registry.unTrackPod(pod)
 		}
+		registry.podTrackingLock.Unlock()
```
pravisankar: Currently this is not an issue but there could be potential issues later on with this approach:
- Since we are taking lock at a wider scope than needed, if a method (trackPod/unTrackPod...) panics for some reason holding the lock, all dependent methods will be blocked
- We need to sprinkle these lock/unlock whenever we access podsByIP

Instead I prefer adding these 3 methods:
GetPodInfo() {
get lock
fetch from podsByIP
release lock
}
SetPodInfo(podInfo) {
get lock
add/set pod info
release lock
}
UnSetPodInfo(podInfo) {
get lock
remove from podsByIP
release lock
}
Use these methods where ever required : in endpointsUpdate, trackPod and untrackPod

danwinship: > Instead I prefer adding these 3 methods:
> ...
> Use these methods where ever required : in endpointsUpdate, trackPod and untrackPod

The lock needs to be held throughout trackPod() and untrackPod() though or else you might test one value and then delete another, etc.

Re-pushed with proper go-ish use of defer to unlock the mutexes, and changed trackPod() to delete stale data itself rather than calling out to unTrackPod() to do it. (unTrackPod() does an additional check which is dropped now, but that check should be irrelevant; if the UID is correct, then we want to remove the pod, regardless of whether the IP matches or not [which it should, but...])

pravisankar:
Looks good

PR #293, opened 2016-04-13
Synchronize and controlled access to vnid map

pravisankar:
Fix for bug# https://bugzilla.redhat.com/show_bug.cgi?id=1323279

### DIFF
--- a/plugins/osdn/vnids.go
+++ b/plugins/osdn/vnids.go
@@ -66,10 +66,19 @@ func (oc *OsdnController) isAdminNamespace(nsName string) bool {
 }
 
 func (oc *OsdnController) assignVNID(namespaceName string) error {
-	_, err := oc.Registry.GetNetNamespace(namespaceName)
-	if err == nil {
+	// Nothing to do if the netid is in the VNIDMap
+	if _, ok := oc.VNIDMap[namespaceName]; ok {
 		return nil
 	}
+
+	// If NetNamespace is present, update VNIDMap
+	netns, err := oc.Registry.GetNetNamespace(namespaceName)
+	if err == nil {
+		oc.VNIDMap[namespaceName] = netns.NetID
+		return nil
+	}
+
+	// NetNamespace not found, so allocate new NetID
 	var netid uint
 	if oc.isAdminNamespace(namespaceName) {
 		netid = AdminVNID
@@ -80,6 +89,8 @@ func (oc *OsdnController) assignVNID(namespaceName string) error {
 			return err
 		}
 	}
+
+	// Create NetNamespace Object and update VNIDMap
 	err = oc.Registry.WriteNetNamespace(namespaceName, netid)
 	if err != nil {
 		e := oc.netIDManager.ReleaseNetID(netid)
--- a/plugins/osdn/vnids.go
+++ b/plugins/osdn/vnids.go
@@ -105,18 +105,24 @@ func (oc *OsdnController) assignVNID(namespaceName string) error {
 }
 
 func (oc *OsdnController) revokeVNID(namespaceName string) error {
+	// Remove NetID from VNIDMap
+	netid, found := oc.VNIDMap[namespaceName]
+	if !found {
+		log.Errorf("Failed to find NetID for namespace: %s in VNIDMap", namespaceName)
+	} else {
+		delete(oc.VNIDMap, namespaceName)
+	}
+
+	// Delete NetNamespace object
 	err := oc.Registry.DeleteNetNamespace(namespaceName)
 	if err != nil {
 		return err
 	}
-	netid, found := oc.VNIDMap[namespaceName]
-	if !found {
-		return fmt.Errorf("Error while fetching Net ID for namespace: %s", namespaceName)
-	}
-	delete(oc.VNIDMap, namespaceName)
 
-	// Skip AdminVNID as it is not part of Net ID allocation
-	if netid == AdminVNID {
+	// Skip NetID release if
+	// - Value matches AdminVNID as it is not part of NetID allocation or
+	// - NetID is not found in the VNIDMap
+	if netid == AdminVNID || !found {
 		return nil
 	}
 
--- a/plugins/osdn/vnids.go
+++ b/plugins/osdn/vnids.go
@@ -10,6 +10,7 @@ import (
 
 	kapi "k8s.io/kubernetes/pkg/api"
 	kubetypes "k8s.io/kubernetes/pkg/kubelet/container"
+	"k8s.io/kubernetes/pkg/util/sets"
 )
 
 const (
@@ -20,11 +21,14 @@ const (
 )
 
 func (oc *OsdnController) GetVNID(name string) (uint, error) {
+	oc.vnidLock.Lock()
+	defer oc.vnidLock.Unlock()
+
 	if id, ok := oc.vnidMap[name]; ok {
 		return id, nil
 	}
 	// In case of error, return some value which is not a valid VNID
-	return MaxVNID + 1, fmt.Errorf("Failed to find NetID for namespace: %s in vnid map", name)
+	return MaxVNID + 1, fmt.Errorf("Failed to find netid for namespace: %s in vnid map", name)
 }
 
 // Nodes asynchronously watch for both NetNamespaces and services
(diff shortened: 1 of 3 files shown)


knobunc:
LGTM

plugins/osdn/vnids.go
```diff
@@ -18,14 +20,104 @@ const (
...
 
+func (oc *OsdnController) GetVNID(name string, retry bool) (uint, error) {
+	oc.vnidLock.Lock()
+	defer oc.vnidLock.Unlock()
+
+	if id, ok := oc.vnidMap[name]; ok {
+		return id, nil
+	}
+
+	// Nodes asynchronously watches for both NetNamespace and service
```
knobunc: Nodes asynchronously watch for...

plugins/osdn/vnids.go
```diff
@@ -18,14 +20,104 @@ const (
...
+	// If for some reason, vnid map propagation from master to node is slow
+	// and if service/pod-setup tries to lookup vnid map then it may fail.
+	// So, this retry option will alleviate this problem.
+	if retry {
+		// Try few times up to 2 seconds
+		retries := 20
+		retryInterval := 100 * time.Millisecond
+		for i := 0; i < retries; i++ {
+			oc.vnidLock.Unlock()
+			time.Sleep(retryInterval)
```
knobunc: Will this conflict with someone else?  Should the wait vary by the number or retries or something so that two don't wake up in sync and then sleep the same length?

pravisankar:
Extended networking tests (test/extended/networking.sh) passed and manually tested join-projects/make-projects-global.

pravisankar:
@openshift/networking ready for review/merge

plugins/osdn/vnids.go
```diff
@@ -105,18 +105,24 @@ func (oc *OsdnController) assignVNID(namespaceName string) error {
...
-		return fmt.Errorf("Error while fetching Net ID for namespace: %s", namespaceName)
-	}
-	delete(oc.VNIDMap, namespaceName)
 
-	// Skip AdminVNID as it is not part of Net ID allocation
-	if netid == AdminVNID {
+	// Skip NetID release if
+	// - Value matches AdminVNID as it is not part of NetID allocation or
+	// - NetID is not found in the VNIDMap
+	if netid == AdminVNID || !found {
```
danwinship: Why do you skip the release if namespaceName was missing from VNIDMap? I mean, I guess this is kind of a "can't happen" so it doesn't matter too much what we do, but if the vnid doesn't appear in the vnidmap, then shouldn't we make sure it's not in use in the netIDManager too?

(Maybe we should merge VNIDMap into netIDManager though... then we wouldn't have to even think about the possibility of them being out of sync.)

plugins/osdn/vnids.go
```diff
@@ -18,14 +20,104 @@ const (
...
+	// NetNamespaces populates vnid map and services/pod-setup depend on vnid map
+	// If for some reason, vnid map propagation from master to node is slow
+	// and if service/pod-setup tries to lookup vnid map then it may fail.
+	// So, this retry option will alleviate this problem.
+	if retry {
+		// Try few times up to 2 seconds
+		retries := 20
+		retryInterval := 100 * time.Millisecond
+		for i := 0; i < retries; i++ {
+			oc.vnidLock.Unlock()
```
danwinship: I'm not sure this is legit; you have a pending `defer oc.vnidLock.Unlock()`... if the time.Sleep() call panicked then you'd end up unlocking the mutex when it was already unlocked. (Which, admittedly would just cause another panic, so it's not like it's the end of the world... but if more code got added to the unlocked section later then this could be a problem.)

plugins/osdn/vnids.go
```diff
@@ -88,44 +185,42 @@ func (oc *OsdnController) assignVNID(namespaceName string) error {
...
-	for name, id := range oc.VNIDMap {
-		if id == netid {
-			netid_inuse = true
-			log.V(5).Infof("Net ID %d for namespace %q is still in use by namespace %q", netid, namespaceName, name)
-			break
-		}
-	}
-	if !netid_inuse {
+	if !oc.checkVNID(netid) {
 		err = oc.netIDManager.ReleaseNetID(netid)
```
danwinship: Are API calls synchronized into a single thread or is it possible for two to be happening at the same time in different goroutines? If the latter, then it's possible that checkVNID() will return false, but then another namespace will be modified to use netid in another thread before the call to ReleaseNetID(). You need to hold the mutex across both calls in order to be safe.

plugins/osdn/vnids.go
pravisankar: There is contention for vnidMap from watchNetNamespaces/watchServices/SetupPod but sleeping for the same length won't be an issue. Even if multiple of them wake up at the same time, only one of them will get the vnidLock and the others will go back to sleep.

plugins/osdn/vnids.go
pravisankar: Okay, I will fix this case.

plugins/osdn/vnids.go
pravisankar: assignVNID/revokeVNID methods are synchronized by the event queue. Even if two threads/goroutines execute at the same time, this may not be an issue. One thread can run checkVNID() and let's say it returned false but the other thread can't use the same netid as the corresponding netnamespace with this netid is already deleted. If there is another netnamespace with the same netid then checkVNID will not return false.

pravisankar:
@openshift/networking Updated, ptal

knobunc:
LGTM

plugins/osdn/vnids.go
```diff
@@ -18,14 +20,103 @@ const (
...
+	return MaxVNID + 1, fmt.Errorf("Failed to find NetID for namespace: %s in vnid map", name)
+}
+
+// Nodes asynchronously watch for both NetNamespaces and services
+// NetNamespaces populates vnid map and services/pod-setup depend on vnid map
+// If for some reason, vnid map propagation from master to node is slow
+// and if service/pod-setup tries to lookup vnid map then it may fail.
+// So, use this method to alleviate this problem. This method will
+// retry vnid lookup and NetNamespaces in etcd before giving up.
+func (oc *OsdnController) TryAndGetVNID(name string) (uint, error) {
```
danwinship: "try" suggests something like `pthread_trylock` to me, which is exactly the opposite of what it actually means here. Maybe `WaitForVNID`?

Also, some of the changes here should be pushed back into the previous commit; it's weird to add a "retry" flag in the previous commit and then remove it in this one and do things differently instead.

pravisankar:
@danwinship Updated, PTAL

plugins/osdn/vnids.go
```diff
@@ -18,14 +20,103 @@ const (
...
+// If for some reason, vnid map propagation from master to node is slow
+// and if service/pod-setup tries to lookup vnid map then it may fail.
+// So, use this method to alleviate this problem. This method will
+// retry vnid lookup and NetNamespaces in etcd before giving up.
+func (oc *OsdnController) WaitAndGetVNID(name string) (uint, error) {
+	// Try few times up to 2 seconds
+	retries := 20
+	retryInterval := 100 * time.Millisecond
+	for i := 0; i < retries; i++ {
+		if id, err := oc.GetVNID(name); err != nil {
```
dcbw: I think this should be "err == nil" ?

plugins/osdn/vnids.go
```diff
@@ -18,14 +20,103 @@ const (
...
+			return id, nil
+		}
+		time.Sleep(retryInterval)
+	}
+
+	// Check if we can get ID from NetNamespace object
+	netns, err := oc.Registry.GetNetNamespace(name)
+	if err == nil {
+		oc.setVNID(name, netns.NetID)
```
dcbw: Also this seems a bit odd; this function gets called from watchServices() and plugin.go::getVNID() only.  It feels like it shouldn't need to update the VNID map at all, since shouldn't that be done already on startup by populateVNIDMap() or at runtime by watchNetNamespaces() and watchNamespaces()-via-assignVNID()?

I feel like this function should only get what's already in the VNID map (possibly waiting for it).  If the ID isn't in the VNID map yet, it'll get added shortly by watchNetNamespaces() or watchNamespaces().  I think centralizing setting the VNID map is less confusing.

dcbw:
@pravisankar added some review comments on the patches.

knobunc:
Don't merge this yet please, we'll hold it for after 3.2.

pravisankar:
@dcbw addressed your comments, thanks.

plugins/osdn/vnids.go
```diff
@@ -88,44 +178,42 @@ func (oc *OsdnController) assignVNID(namespaceName string) error {
...
 		err = oc.netIDManager.ReleaseNetID(netid)
 		if err != nil {
-			return fmt.Errorf("Error while releasing Net ID: %v", err)
-		} else {
-			log.Infof("Released netid %d for namespace %q", netid, namespaceName)
+			return fmt.Errorf("Error while releasing netid %d for namespace %q, %v", netid, namespaceName, err)
 		}
+		log.Infof("Released netid %d for namespace %q", netid, namespaceName)
+	} else {
+		log.V(5).Infof("Net ID %d for namespace %q is still in use", netid, namespaceName)
```
knobunc: Can we use either "net id" or "netid" consistently in the log messages to make it easier to grep for?

dcbw:
LGTM now

plugins/osdn/vnids.go
pravisankar: done

danwinship:
LGTM too... I'm not sure where we are with respect to freezes now though

PR #274, opened 2016-03-24
Removed the cookies from the remote node rules.

knobunc:
The rest of the ovs sdn rules had removed them in other patches.  This catches the other cases we hadn't yet changed.

PR #294, opened 2016-04-13
Fix AddHostSubnetRules() and DeleteHostSubnetRules()

pravisankar:


pravisankar:
We missed these fixes in https://github.com/openshift/openshift-sdn/pull/274
@knobunc @danwinship @dcbw  PTAL

dcbw:
LGTM

PR #295, opened 2016-04-14
Return better errors from SetUpPod/TearDownPod

### LINKED ISSUE #290: When SDN is not configured properly, insufficient debug info printed to logs
This may be a Kube issue (if Kube is eating necessary debug info) but starting here.  When running with multi tenant plugin with some system configuration not correct, the only message the Kubelet returns on startup is:

```
E0407 01:05:44.976884   16746 manager.go:1871] Failed to setup network for pod "docker-registry-1-kwj8m_default(5aefb286-fb0f-11e5-83ac-080027c5bfa9)" using network plugins "redhat/openshift-ovs-multitenant": exit status 1; Skipping pod
```

That means that it's impossible for an admin to take the next step and debug why the configuration is inaccurate.  In general, whenever we fail in a flow we should print out necessary info to debug further (I assume in this case stdout/stderr from the execution of the network plugin script).

danwinship:
Fixes #290 
@openshift/networking PTAL

### DIFF
--- a/plugins/osdn/ovs/plugin.go
+++ b/plugins/osdn/ovs/plugin.go
@@ -3,7 +3,9 @@ package ovs
 import (
 	"fmt"
 	"net"
+	"os/exec"
 	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 
@@ -14,7 +16,6 @@ import (
 	"k8s.io/kubernetes/pkg/api/resource"
 	kubeletTypes "k8s.io/kubernetes/pkg/kubelet/container"
 	knetwork "k8s.io/kubernetes/pkg/kubelet/network"
-	utilexec "k8s.io/kubernetes/pkg/util/exec"
 )
 
 type ovsPlugin struct {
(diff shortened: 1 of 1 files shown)


plugins/osdn/ovs/plugin.go
```diff
@@ -210,16 +227,26 @@ func (plugin *ovsPlugin) SetUpPod(namespace string, name string, id kubeletTypes
 		return err
 	}
 
-	out, err := utilexec.New().Command(plugin.getExecutable(), setUpCmd, string(id), vnidstr, ingressStr, egressStr, fmt.Sprintf("%t", macvlan)).CombinedOutput()
+	out, err := exec.Command(plugin.getExecutable(), setUpCmd, string(id), vnidstr, ingressStr, egressStr, fmt.Sprintf("%t", macvlan)).CombinedOutput()
 	glog.V(5).Infof("SetUpPod network plugin output: %s, %v", string(out), err)
-	return err
+
+	if isScriptError(err) {
+		return fmt.Errorf("Error running network setup script: %s", getScriptError(out))
```
pravisankar: Can you also include pod name and namespace in the error msg? I think that will be useful.

pravisankar:
Minor comment, LGTM overall

plugins/osdn/ovs/plugin.go
danwinship: kubernetes adds that when it logs the error:

    message := fmt.Sprintf("Failed to setup network for pod %q using network plugins %q: %v; Skipping pod", format.Pod(pod), dm.networkPlugin.Name(), err)

where `format.Pod(pod)` outputs `fmt.Sprintf("%s_%s(%s)", pod.Name, pod.Namespace, pod.UID)`

dcbw:
LGTM

PR #296, opened 2016-04-20
Add more HostSubnet logging

dcbw:
We don't expect nodes to be added/removed very often so the log
spam won't be large.

### DIFF
--- a/plugins/osdn/common.go
+++ b/plugins/osdn/common.go
@@ -51,6 +51,8 @@ type OsdnController struct {
 // Called by plug factory functions to initialize the generic plugin instance
 func (oc *OsdnController) BaseInit(registry *Registry, pluginHooks PluginHooks, multitenant bool, hostname string, selfIP string) error {
 
+	log.Infof("Starting with configured hostname '%s' (IP '%s')", hostname, selfIP)
+
 	if hostname == "" {
 		output, err := kexec.New().Command("uname", "-n").CombinedOutput()
 		if err != nil {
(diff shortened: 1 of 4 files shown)


plugins/osdn/common.go
```diff
@@ -51,6 +51,8 @@ type OsdnController struct {
 // Called by plug factory functions to initialize the generic plugin instance
 func (oc *OsdnController) BaseInit(registry *Registry, pluginHooks PluginHooks, multitenant bool, hostname string, selfIP string) error {
 
+	log.Infof("Starting with configured hostname '%s' (IP '%s')", hostname, selfIP)
+
```
pravisankar: We are already logging hostname/selfIP in https://github.com/dcbw/openshift-sdn/blob/more-hostsubnet-logging/plugins/osdn/common.go#L77-L79

dcbw: Yeah, but that's after the fixups.  I'd like to see what the configured/command-line values are too.

plugins/osdn/subnets.go
```diff
@@ -49,7 +51,12 @@ func (oc *OsdnController) addNode(nodeName string, nodeIP string) error {
 		return fmt.Errorf("Error writing subnet %v to etcd for node %s: %v", sn, nodeName, err)
 	}
 
-	log.Infof("Created HostSubnet %v for node %s (%s)", sn, nodeName, nodeIP)
+	sub, err := oc.Registry.GetSubnet(nodeName)
+	if err != nil {
+		return fmt.Errorf("Error fetching subnet for node %q: %v", nodeName, err)
+	}
```
pravisankar: Unnecessary etcd call just for logging. Instead we could make CreateSubnet() to return the subnet along with the error (currently we are ignoring the returned subnet: https://github.com/dcbw/openshift-sdn/blob/more-hostsubnet-logging/plugins/osdn/registry.go#L105)

pravisankar:
LGTM

PR #300, opened 2016-04-25
Split proxy endpoint filtering out into separate object

dcbw:
With more origin code this should give us the ability to run the proxy in a separate process.

I chose to keep the registry since the proxy also needs the origin client to get the ClusterNetwork, but I could split that out and simplify things a but further if anyone wants?

### DIFF
--- a/plugins/osdn/factory/factory.go
+++ b/plugins/osdn/factory/factory.go
@@ -12,8 +12,17 @@ import (
 	"github.com/openshift/openshift-sdn/plugins/osdn/ovs"
 )
 
-// Call by higher layers to create the plugin instance
-func NewPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, error) {
+// Call by higher layers to create the plugin SDN master instance
+func NewMasterPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client) (api.OsdnPlugin, error) {
+	return newPlugin(pluginType, osClient, kClient, "", "")
+}
+
+// Call by higher layers to create the plugin SDN node instance
+func NewNodePlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, error) {
+	return newPlugin(pluginType, osClient, kClient, hostname, selfIP)
+}
+
+func newPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, error) {
 	switch strings.ToLower(pluginType) {
 	case ovs.SingleTenantPluginName:
 		return ovs.CreatePlugin(osdn.NewRegistry(osClient, kClient), false, hostname, selfIP)
--- a/plugins/osdn/factory/factory.go
+++ b/plugins/osdn/factory/factory.go
@@ -13,7 +13,7 @@ import (
 )
 
 // Call by higher layers to create the plugin instance
-func NewPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, api.FilteringEndpointsConfigHandler, error) {
+func NewPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, error) {
 	switch strings.ToLower(pluginType) {
 	case ovs.SingleTenantPluginName:
 		return ovs.CreatePlugin(osdn.NewRegistry(osClient, kClient), false, hostname, selfIP)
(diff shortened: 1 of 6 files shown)


dcbw:
origin side is https://github.com/openshift/origin/pull/8627

danwinship:
Will need to be rebased on top of #289 (particularly the commit "Simplify Watch resource in SDN")

plugins/osdn/factory/factory.go
```diff
@@ -13,13 +13,23 @@ import (
 )
 
 // Call by higher layers to create the plugin instance
-func NewPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, api.FilteringEndpointsConfigHandler, error) {
+func NewPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, error) {
```
danwinship: if we're going to change the API here, then can we also split this into NewMasterPlugin() and NewNodePlugin()? (Where for now `NewMasterPlugin(pluginType, osClient, kClient)` just calls `NewNodePlugin(pluginType, osClient, kClient, "", "")`). Yeah, it doesn't have anything to do with the rest of this patch, but if we're going to be having to do synced origin/sdn changes anyway...

dcbw: sure

plugins/osdn/ovs/proxy.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Only delete if the pod ID is the one we are tracking (in case there is a failed or complete
+	// pod lying around that gets deleted while there is a running pod with the same IP)
+	if podInfo, ok := proxy.podsByIP[pod.Status.PodIP]; ok && podInfo.UID == pod.UID {
+		delete(proxy.podsByIP, pod.Status.PodIP)
+	}
+}
+
+// FilteringEndpointsConfigHandler implementation
+func (proxy *ovsProxyPlugin) SetBaseEndpointsHandler(base pconfig.EndpointsConfigHandler) {
```
pravisankar: Can we pass `base pconfig.EndpointsConfigHandler` to CreateProxyPlugin and get rid of this method?

plugins/osdn/ovs/proxy.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/openshift/openshift-sdn/plugins/osdn/api"
+
+	kapi "k8s.io/kubernetes/pkg/api"
+	pconfig "k8s.io/kubernetes/pkg/proxy/config"
+)
+
+type ovsProxyPlugin struct {
+	registry *osdn.Registry
+	podsByIP map[string]*kapi.Pod
+	mutex    sync.Mutex
```
pravisankar: Generic name `mutex` may give the impression that it protects the whole struct. Can we rename mutex so that we know it's only needed for podsByIP map.

plugins/osdn/factory/factory.go
pravisankar: NewMasterPlugin() calling NewNodePlugin() looks weird. Instead add new wrapper methods NewMaserPlugin(), NewNodePlugin() and both of them can call NewPlugin/NewOsdnPlugin() with approriate params.

plugins/osdn/ovs/proxy.go
pravisankar: Looking at the origin pr, we don't have pconfig.EndpointsConfigHandler when we call CreateProxyPlugin(), can be moved to Start() as Dan Winship suggested?

dcbw:
@pravisankar changed the name

danwinship:
You merged your fixes into the wrong commit. Also, I think you missed Ravi's request to have NewMasterPlugin() and NewNodePlugin() both call a private method rather than having NewMasterPlugin() calling NewNodePlugin() (which, admittedly, was a lame suggestion on my part :)

dcbw:
@danwinship right on all counts; fixed now.

danwinship:
LGTM

dcbw:
@pravisankar @rajatchopra

plugins/osdn/registry.go
```diff
@@ -81,16 +73,13 @@ func (registry *Registry) CreateSubnet(nodeName, nodeIP, subnetCIDR string) (*os
 	return registry.oClient.HostSubnets().Create(hs)
 }
 
-func (registry *Registry) PopulatePodsByIP() error {
+func (registry *Registry) GetAllPods() ([]kapi.Pod, error) {
 	podList, err := registry.kClient.Pods(kapi.NamespaceAll).List(kapi.ListOptions{})
 	if err != nil {
-		return err
+		return make([]kapi.Pod, 0), err
```
pravisankar: return nil, err ?

pravisankar:
minor nit, LGTM

dcbw:
@pravisankar nit fixed

PR #301, opened 2016-04-26
Refer SDN plugin names and annotations by constants

pravisankar:
Remove unused GetName() from ovs controller and we already have
Name() to return the name of the sdn plugin if needed.


pravisankar:
@openshift/networking PTAL

danwinship:
LGTM

dcbw:
LGTM

PR #302, opened 2016-04-28
Ignore malformed IP in Node object

danwinship:
ought to fix https://bugzilla.redhat.com/show_bug.cgi?id=1320959

@openshift/networking PTAL


pravisankar:
LGTM

dcbw:
I feel like we should warn when this happens, or at least get clarification from upstream whether node.Status.Addresses is actually supposed to be filled in.

eparis:
I feel like there was something about dropping support for one of those node ip fields...  Although I think that was in pod now that I think about it...

PR #303, opened 2016-04-29
Support network ID allocation as annotation on namespace resource

pravisankar:
Trello card: https://trello.com/c/nULFSjXm/29-13-store-vnid-as-annotation-on-namespace-techdebt-supportability


pkg/netid/netid.go
```diff
@@ -0,0 +1,61 @@
...
+	"strconv"
+
+	kapi "k8s.io/kubernetes/pkg/api"
+)
+
+const (
+	// Current assigned VNID for the namespace
+	VNIDAnnotation string = "pod.network.openshift.io/multitenant.vnid"
+	// Desired VNID for the namespace
+	WantsVNIDAnnotation string = "pod.network.openshift.io/multitenant.wants-vnid"
```
danwinship: Clayton's argument for having "pod" in the name of the macvlan annotation was:

> You want to make sure that you identify that this is related to pods, so "pod" needs to be in the name somewhere.

So "pod" seems wrong here... Maybe "cluster"?

pkg/netid/netid.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var (
+	ErrorVNIDNotFound = fmt.Errorf("VNID/WantsVNID annotation not found")
```
danwinship: given that annotation names themselves tend to have slashes in them, it would be better to not use one as a non-name-component in the error message text. ("VNID or WantsVNID" would be a simple fix, but you could also just create the error when you throw it and %s the annotationKey value into it...)

plugins/osdn/vnids.go
```diff
@@ -72,194 +111,37 @@ func (oc *OsdnController) unSetVNID(name string) (id uint, err error) {
...
-		return nil
-	}
-
-	// Check if this netid is used by any other namespaces
-	// If not, then release the netid
-	if !oc.checkVNID(netid) {
-		err = oc.netIDManager.ReleaseNetID(netid)
-		if err != nil {
-			return fmt.Errorf("Error while releasing netid %d for namespace %q, %v", netid, namespaceName, err)
+			continue
```
danwinship: since I'm guessing this will show a useless code snippet, I'm commenting on:

    func populateVNIDMap(oc *OsdnController) error {
        ...
        id, err := netid.GetVNID(&ns)
        if err != nil { 
            continue

This ignores both "annotation isn't set" and "annotation is set, but to a nonsense-value", which we probably ought to warn about

plugins/osdn/netid/vnidallocator/allocator.go
```diff
@@ -0,0 +1,145 @@
...
+
+// contains returns true and the offset if the VNID is in the range,
+// and false and 0 otherwise.
+func (r *Allocator) contains(id uint) (bool, uint) {
+	if !r.vnidRange.Contains(id) {
+		return false, 0
+	}
+
+	offset := id - r.vnidRange.Base
+	return true, offset
```
danwinship: random that you call a VNIDRange method for the first part, but then do the second part by hand

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,169 @@
...
+// retryCount is the number of times to retry on a conflict when updating a namespace
+const retryCount = 2
+
+// AddOrUpdate allocates network id(VNID) for the namespace if needed
+func (c *Allocation) AddOrUpdate(ns *kapi.Namespace) error {
+	tx := &tx{}
+	defer tx.Rollback()
+
+	var id uint
+	vnidAllocated := false
```
danwinship: vnidAllocated isn't a good name for this variable. (userProvidedVnid?)

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,169 @@
...
+		// Skip current namespace
+		if n.ObjectMeta.Name == name {
+			continue
+		}
+		// Skip namespaces that doesn't match given id
+		if value, err := netid.GetVNID(&n); (err != nil) || (value != id) {
+			continue
+		}
+		// VNID is in use by other namespace
+		glog.Infof("Ignored releasing netid %d for namespace %q (already in use by %q)", id, name, n.ObjectMeta.Name)
```
danwinship: "still", not "already"

plugins/osdn/netid/controller/factory.go
```diff
@@ -0,0 +1,76 @@
...
+	utilwait "k8s.io/kubernetes/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/watch"
+
+	"github.com/openshift/openshift-sdn/plugins/osdn/netid/vnidallocator"
+	"github.com/openshift/origin/pkg/client/cache"
+	"github.com/openshift/origin/pkg/controller"
+)
+
+// AllocationFactory can create an Allocation controller.
+type AllocationFactory struct {
```
danwinship: shouldn't there be a constructor for this?

pkg/netid/netid.go
pravisankar: My intent was to convey this annotation is for pod network in openshift and it holds VNID when multitenant network plugin is used. 
'cluster' doesn't sound right...
@smarterclayton do you have any suggestions?

plugins/osdn/netid/vnidallocator/allocator.go
pravisankar: sorry I didn't understand, could you elaborate?

pkg/netid/netid.go
smarterclayton: Quick question - why is there a "wants-vnid" and "vnid"?  When is wants != vnid?

smarterclayton: In this case the vnid would belong to the namespace, but it's related to the pod network still.  Is the VNID specific only to multi tenant plugin, or can there be other plugins that would use this annotation?

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,169 @@
...
+type tx struct {
+	rollback []func() error
+}
+
+func (tx *tx) Add(fn func() error) {
+	tx.rollback = append(tx.rollback, fn)
+}
+
+func (tx *tx) Rollback() {
+	for _, fn := range tx.rollback {
```
smarterclayton: Why does this construct exist?

plugins/osdn/netid/vnidallocator/allocator.go
danwinship: It's just a weird mix of abstraction and not. If you're willing to look directly at vnidRange.Base, then why not look at vnidRange.Size too rather than calling vnidRange.Contains()? Or, if you think you ought to call vnidRange.Contains() for the first part, then there should be a vnidRange.GetOffset() for the second part. (Or vnidRange.Contains() should return the multiple values itself)

pkg/netid/netid.go
pravisankar: > why is there a "wants-vnid" and "vnid"? When is wants != vnid?
VNIDAnnotation represents the current network ID for the namespace and we want this annotation to be assigned only by the vnid controller (watches on namespace changes) to avoid any race conditions.
WantsVNIDAnnotation is intended to be used for requesting a particular vnid (example: oadm pod-network join-projects/make-projects-global). Vnid controller may satisfy this request if certain conditions are met (example: requested id already allocated, within defined range, etc.).

pravisankar: > Is the VNID specific only to multi tenant plugin?

Any OVS based networking solution can use this annotation. In our case, it's only used by multi-tenant network plugin.

plugins/osdn/netid/controller/controller.go
pravisankar: This is used in AddOrUpdate(): we reserve/allocate vnid and then persists this vnid as annotation on the namespace. If we fail to persist the namespace object, as part of rollback we release the allocated vnid.

danwinship:
It belatedly occurs to me that this PR moves us *away* from the upstream NetworkPolicy proposal. In particular, the upstream model allows for multiple independently-isolated networks in a single Namespace, which doesn't seem compatible with having a single VNID annotation per Namespace.

If instead we keep NetNamespaces around, we can think of them as being proto-NetworkPolicies: having a NetNamespace object in a namespace is equivalent to having a collection of NetworkPolicy objects in the namespace that describe an OpenShift-3.2-compatible isolation policy. (Roughly: one NetworkPolicy saying that all pods in the namespace accept traffic from all other pods in the namespace, and a second NetworkPolicy listing each namespace that has the same VNID as this one, and saying that all pods in this namespace accept traffic from all of those namespaces.)

We can still use the controller/annotation changes from this PR though: just have the controller create the NetNamespace objects and assign their VNIDs, and have the WantsVNID annotation be used on the NetNamespace rather than the Namespace... (We could even keep the VNID as an annotation on the NetNamespace rather than being a "NetID" field like now, in preparation for the eventual move to NetworkPolicies, where we'd have to use an annotation for the VNID since there isn't a field.)

danwinship:
@dcbw ^^^

pkg/netid/netid.go
smarterclayton: Are you proposing to add a new virtual resource that allows unprivileged admins to request VNIDS?  Because having these split doesn't add any security, and is basically inconsistent with our other admin commands.  Unless you have a proposal to have separate security models, we should just use `vnid` (that's consistent with every other allocator, like unix users, selinux labels, fsgroup, etc).

pravisankar: Currently we do *not* allow unprivileged admins to request VNID.
The use cases addressed by vnid allocator is slightly different from other existing allocators.
(1) Assign/Release VNID during project creation/deletion respectively:
     vnid controller performs this task (similar to other allocators)
(2) Allow admin to manipulate VNIDs using oadm cmd:
     Some vnids are shared by multiple namespaces and release is done only when no namespace needs that vnid.
     If we allow admin cli to directly manipulate the vnid annotation. We could end up in a situation where admin CLI patches vnid in a namespace but the same vnid could be released by controller at the same time. We can't provide synchronization here because only vnid controller knows which ids are allocated. wants-vnid annotation provides this synchronization. Admin CLI requests needed vnid as wants-vnid annotation and vnid controller verifies if the requested vnid is allowed or not and updates vnid annotation. wants-vnid annotation is removed after every request (either success or failure). Eventually the idea is to replace wants-vnid annotation with network-policy objects (once upstream implements https://github.com/kubernetes/kubernetes/pull/24154).

pravisankar:
On Fri, May 6, 2016 at 9:14 AM, Dan Winship <notifications@github.com>
wrote:

> It belatedly occurs to me that this PR moves us *away* from the upstream
> NetworkPolicy proposal. In particular, the upstream model allows for
> multiple independently-isolated networks in a single Namespace, which
> doesn't seem compatible with having a single VNID annotation per Namespace.
>
> If instead we keep NetNamespaces around, we can think of them as being
> proto-NetworkPolicies: having a NetNamespace object in a namespace is
> equivalent to having a collection of NetworkPolicy objects in the namespace
> that describe an OpenShift-3.2-compatible isolation policy. (Roughly: one
> NetworkPolicy saying that all pods in the namespace accept traffic from all
> other pods in the namespace, and a second NetworkPolicy listing each
> namespace that has the same VNID as this one, and saying that all pods in
> this namespace accept traffic from all of those namespaces.)
>
> We can still use the controller/annotation changes from this PR though:
> just have the controller create the NetNamespace objects and assign their
> VNIDs, and have the WantsVNID annotation be used on the NetNamespace rather
> than the Namespace... (We could even keep the VNID as an annotation on the
> NetNamespace rather than being a "NetID" field like now, in preparation for
> the eventual move to NetworkPolicies, where we'd have to use an annotation
> for the VNID since there isn't a field.)
>

We have decided to tackle this in a separate PR

pravisankar:
@openshift/networking Updated, PTAL

danwinship:
I guess I don't have a strong sense of go code organization practices, but is there any reason for splitting it up so much rather than just have a single plugins/osdn/vnid/ directory containing all of the new code? The vnidallocator is only used by the controller, so why not have it be part of that module? And why is the migration code in a separate module from the rest of the controller when the repair code isn't?

Also, should pkg/netid/ be pkg/vnid/ ?

pkg/cmd/admin/network/isolate_projects.go
```diff
@@ -62,16 +63,17 @@ func NewCmdIsolateProjectsNetwork(commandName, fullName string, f *clientcmd.Fac
...
 	if err != nil {
 		return err
 	}
 
 	errList := []error{}
-	for _, project := range projects {
-		// TBD: Create or Update network namespace
+	for _, info := range namespacesInfo {
+		// TBD: Update/Patch namespace
 		// TODO: Fix this once we move VNID allocation to REST layer
```
danwinship: Are the TDB/TODO comments still relevant? Is the PR incomplete?

plugins/osdn/netid/migration/migration.go
```diff
@@ -0,0 +1,89 @@
...
+			m.deleteNetNamespace(&netns)
+			continue
+		}
+		if err != nil {
+			return err
+		}
+		// Delete processed NetNamespace object
+		m.deleteNetNamespace(&netns)
+
+		// Validate
```
danwinship: don't delete the NetNamespace until after validating?

plugins/osdn/factory/factory.go
```diff
@@ -3,31 +3,31 @@ package factory
...
 
+	"github.com/openshift/openshift-sdn/plugins/osdn"
+	"github.com/openshift/openshift-sdn/plugins/osdn/api"
 	"github.com/openshift/openshift-sdn/plugins/osdn/ovs"
 )
 
 // Call by higher layers to create the plugin SDN master instance
-func NewMasterPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client) (api.OsdnPlugin, error) {
-	return newPlugin(pluginType, osClient, kClient, "", "")
+func NewMasterPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, etcdHelper storage.Interface) (api.OsdnPlugin, error) {
```
danwinship: There's no way to extract the `etcdHelper` from `osClient` or `kClient`? Or vice versa? They're all basically just different APIs around the same underlying database handle, right? There ought to be some better way to do this...

liggitt: no, the clients speak to the API, not to etcd.

danwinship:
It would be nice to get an LGTM on the controller parts from someone who is familiar with the existing controller stuff

plugins/osdn/factory/factory.go
liggitt: Why does the plugin need etcd access? I would have expected it to work against the API

danwinship: the migration code needs to read/delete the existing NetNamespace objects, and presumably Ravi is planning to delete the parts of origin/pkg/client that handle those

plugins/osdn/netid/controller/factory.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// GlobalNamespaces are the namespaces that have access to all pods in the cluster and vice versa.
+	GlobalNamespaces []string
+}
+
+func NewAllocationFactory(vnida vnidallocator.Interface, client kclient.NamespaceInterface, globalNamespaces []string) *AllocationFactory {
+	return &AllocationFactory{
+		VNIDAllocator:    vnida,
+		Client:           client,
+		GlobalNamespaces: globalNamespaces,
```
smarterclayton: How big is this?

plugins/osdn/ovs/plugin.go
```diff
@@ -54,8 +55,13 @@ func (plugin *ovsPlugin) PluginStartMaster(clusterNetwork *net.IPNet, hostSubnet
 		if err := plugin.VnidStartMaster(); err != nil {
 			return err
 		}
+	} else {
+		// Network plugin may be switched from multi-tenant to single-tenant
+		// Delete VNID annotations on the namespaces if needed
+		if err := plugin.VnidCleanup(); err != nil {
```
smarterclayton: This would cause data loss in the event of accidental configuration.  You don't need to do this.

smarterclayton: An admin who wants to fix this up can run `oc annotate VNIDANNOTATION- namespaces --all` if they really want

plugins/osdn/netid/migration/migration.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		// Update namespace
+		updatedNs, err := m.client.Update(ns)
+		if errors.IsNotFound(err) {
+			m.deleteNetNamespace(&netns)
+			continue
+		}
+		if err != nil {
+			return err
```
smarterclayton: you need to tell administrators what they need to do if you're going to error out here.

plugins/osdn/netid/migration/migration.go
```diff
@@ -0,0 +1,89 @@
...
+
+	for _, netns := range netnsList.Items {
+		// Get corresponding namespace for the NetNamespace object
+		ns, err := m.client.Get(netns.NetName)
+		if errors.IsNotFound(err) {
+			m.deleteNetNamespace(&netns)
+			continue
+		}
+		if err != nil {
+			return fmt.Errorf("Get namespace %q failed: %v", netns.NetName, err)
```
smarterclayton: all of the errors in this function need to explain to the admin what is going on and what is actionable.  None of these are good enough to debug a migration in progress.

plugins/osdn/netid/migration/migration.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		} else {
+			return fmt.Errorf("Failed to migrate netid %d from NetNamespace to annotation on namespace %q", netns.NetID, ns.Name)
+		}
+	}
+	return nil
+}
+
+func (m *Migrate) deleteNetNamespace(netns *api.NetNamespace) {
+	var out api.NetNamespace
```
smarterclayton: Why is this not a client API call?

plugins/osdn/netid/migration/migration.go
```diff
@@ -0,0 +1,89 @@
...
+	return &Migrate{
+		etcdHelper: etcdHelper,
+		client:     client,
+	}
+}
+
+func (m *Migrate) Run() error {
+	var netnsList api.NetNamespaceList
+
+	err := m.etcdHelper.List(context.TODO(), etcdNetNamespacePrefix, "", storage.Everything, &netnsList)
```
smarterclayton: Why is this an etcd call?

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,170 @@
...
+		// VNID is in use by other namespace
+		glog.Infof("Ignored releasing netid %d for namespace %q (still in use by %q)", id, name, n.ObjectMeta.Name)
+		return nil
+	}
+
+	err = c.vnid.Release(id)
+	if err != nil {
+		return err
+	}
+	glog.Infof("Released netid %d for namespace %q", id, name)
```
smarterclayton: Don't glog.Infof - these should be at least glog.v(4).

pravisankar:
On Wed, May 11, 2016 at 1:33 PM, Dan Winship <notifications@github.com>
wrote:

> I guess I don't have a strong sense of go code organization practices, but
> is there any reason for splitting it up so much rather than just have a
> single plugins/osdn/vnid/ directory containing all of the new code? The
> vnidallocator is only used by the controller, so why not have it be part of
> that module? And why is the migration code in a separate module from the
> rest of the controller when the repair code isn't?
>
Code was organized similar to other existing allocators in k8s and origin.
Controller dir contains both controllers (vnid and repair) and migration is
not included because it's for netnamespace resource.

> Also, should pkg/netid/ be pkg/vnid/ ?
>
I chose netid (network ID) as the top level dir (since vnid is used for
vnidRange interface). So, I kept common stuff in CLI and OVS in pkg/netid
and OVS specific stuff under plugins/osdn/netid.

pkg/cmd/admin/network/isolate_projects.go
pravisankar: Stale TBD/TODO, I will remove.

plugins/osdn/factory/factory.go
pravisankar: We can't directly use the API in this case. As the existing NetNamespace resource is converted to virtual resource backed by namespace to support backward compatibility, we need the etcd access to directly access /registry/sdnnetnamespaces for migration.

plugins/osdn/netid/migration/migration.go
pravisankar: We don't have direct access to /registry/sdnnetnamespaces via client/rest api. 
Now, NetNamespace is a virtual resource backed by namespace (to support old clients).

plugins/osdn/factory/factory.go
liggitt: entangling the plugin with etcd and migration seems less than ideal. Shouldn't the migration responsibility live where the registry was previously, and the plugin should continue interacting only with the API?

plugins/osdn/netid/migration/migration.go
liggitt: Virtual resources are accessible via the API. That's sort of the definition of them... they are API resources with no backing storage (they're backed by something else, in this case, namespace annotations)

plugins/osdn/netid/controller/controller.go
pravisankar: This is helpful in sdn troubleshooting. Initially we added these logs at v(5) but these were not logged as users are not running with --loglevel=5 most of the time. We get one log entry for every project add or delete, so we converted to glog.Infof

plugins/osdn/netid/controller/factory.go
pravisankar: currently, we only got one global namespace: 'default'

plugins/osdn/netid/migration/migration.go
pravisankar: Yes, Virtual resources are accessible via API but none of the create/update/delete apis hits old persistent etcd store (/registry/sdnnetnamespaces)?

plugins/osdn/factory/factory.go
pravisankar: I also need ectd helper for vnid ranges: https://github.com/openshift/openshift-sdn/pull/303/commits/08ae21eaca31a2e10e27845f111efa0016d7f24a#diff-643475b06fe7640bd0eab5a38497f7beR34

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,170 @@
...
+	"k8s.io/kubernetes/pkg/api/errors"
+	kclient "k8s.io/kubernetes/pkg/client/unversioned"
+	utilruntime "k8s.io/kubernetes/pkg/util/runtime"
+	"k8s.io/kubernetes/pkg/util/sets"
+
+	"github.com/openshift/openshift-sdn/pkg/netid"
+	"github.com/openshift/openshift-sdn/plugins/osdn/netid/vnidallocator"
+)
+
+type Allocation struct {
```
dcbw: How about NetidController or VnidController?

plugins/osdn/netid/controller/controller.go
dcbw: Yeah, we continually run into deployments we need to debug where there's simply not enough logging to figure out what's going on.  We can't really ask the users to restart their cluster and reproduce the problem.  So we've been steadily adding logging to make it somewhat possible to debug issues.  We know we don't want logspam, so we're limiting the Info/v2/v3 stuff to things that happen rarely, like node add/remove, namespace add/remove, etc.

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,170 @@
...
+			} else {
+				glog.Infof("Assigned netid %d for namespace %q", id, name)
+			}
+			// commit and exit
+			tx.Commit()
+			return nil
+		}
+
+		if errors.IsNotFound(err) {
+			return nil
```
dcbw: When will IsNotFound() happen?  Shouldn't this return some kind of error instead of nil?

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		if !errors.IsConflict(err) {
+			return err
+		}
+		newNs, err := c.client.Get(name)
+		if errors.IsNotFound(err) {
```
dcbw: Here too; maybe just a comment indicating what a nil return means for this function?

plugins/osdn/netid/controller/factory.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// handleOne processes resource with Handle.
+func (f *AllocationFactory) handleOne(eventType watch.EventType, resource interface{}, err error) {
+	if err != nil {
+		glog.Errorf("Failed to pop namespace from event queue: %v", err)
+		return
+	}
+
+	c := &Allocation{
```
dcbw: Maybe Allocation should just get folded into factory.go?  Neither one is that large and they both use the same data.

plugins/osdn/netid/controller/controller.go
pravisankar: When we try to get/update the namespace and if the namespace got deleted (async by some other process), then IsNotFound(err) returns true. In this case, we don't have to patch anything so we don't return an error.

pravisankar:
@openshift/networking Updated, PTAL

plugins/osdn/netid/vnid/vnid.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/openshift/openshift-sdn/pkg/netid"
+)
+
+type VNIDRange struct {
+	Base uint
+	Size uint
+}
+
+// Contains tests whether a given vnid falls within the Range.
+func (r *VNIDRange) Contains(vnid uint) (bool, uint) {
```
dcbw: Could you add something to the code doc about what 'offset' is and why it gets used on return, instead of using the passed in vnid directly?

dcbw: eh, nevermind; it's because the k8s allocators want it that way

knobunc:
LGTM

@liggitt @smarterclayton Have your concerns been addressed by the latest version of the PR?

pkg/cmd/admin/network/project_options.go
```diff
@@ -4,40 +4,32 @@ import (
...
-	osclient "github.com/openshift/origin/pkg/client"
+	"github.com/openshift/openshift-sdn/pkg/netid"
 	"github.com/openshift/origin/pkg/cmd/util/clientcmd"
-	"github.com/openshift/origin/pkg/project/api"
-	sdnapi "github.com/openshift/origin/pkg/sdn/api"
 )
 
 const (
-	ovsPluginName = "redhat/openshift-ovs-multitenant"
+	ovsPluginName string = "redhat/openshift-ovs-multitenant"
```
liggitt: type shouldn't be necessary

pkg/cmd/admin/network/project_options.go
```diff
@@ -109,83 +97,82 @@ func (p *ProjectOptions) GetProjects() ([]*api.Project, error) {
...
-			return projectList, fmt.Errorf("Projects %v not found", strings.Join(skippedProjectNames.List(), ", "))
+			return infoList, fmt.Errorf("Projects %v not found", strings.Join(skippedProjectNames.List(), ", "))
 		}
 	}
-	return projectList, nil
+	return infoList, nil
 }
 
-func (p *ProjectOptions) GetNetNamespaces() (*sdnapi.NetNamespaceList, error) {
-	netNamespaces, err := p.Oclient.NetNamespaces().List(kapi.ListOptions{})
```
liggitt: does this mean this command won't work against older servers?

pravisankar: Yes, new client with old server will not work. Do we need this?
We want to obsolete NetNamespace resource at some point. We are supporting oadm pod-network command to work with both old/new clients against new servers.

liggitt: we do a lot of other work to make new clients work with old servers, at least for N versions. @smarterclayton, what were you expecting here?

smarterclayton: I think for this specific case, given that it is an admin level task and a
specific command that is infrequently called, and the goal is to simplify
and strip down these APIs more generally, that we can drop the support.
However, we should at minimum handle the error.

liggitt: I'm more concerned that there will be no error (this is just annotating a namespace, which will work against older servers, but not actually accomplish isolation)

pkg/netid/netid.go
```diff
@@ -0,0 +1,94 @@
...
+	"strconv"
+
+	kapi "k8s.io/kubernetes/pkg/api"
+)
+
+const (
+	// Maximum VXLAN Virtual Network Identifier(VNID) as per RFC#7348
+	MaxVNID = uint((1 << 24) - 1)
+	// VNID: 1 to 9 are internally reserved for any special cases in the future
+	MinVNID = 10
```
liggitt: is this a new restriction? what happens if there are existing vinds in this range? Do we have a good sense what we'll use the "special" ones for, and is 10 enough?

pkg/netid/netid.go
```diff
@@ -0,0 +1,94 @@
...
+	// VNID: 0 reserved for default namespace and can reach any network in the cluster
+	GlobalVNID = uint(0)
+
+	// Current assigned VNID for the namespace
+	VNIDAnnotation string = "pod.network.openshift.io/multitenant.vnid"
+	// Desired VNID for the namespace
+	WantsVNIDAnnotation string = "pod.network.openshift.io/multitenant.wants-vnid"
```
liggitt: from the name, I'd expect this to be true/false. would `pod.network.openshift.io/multitenant.requested-vnid` be more descriptive?

plugins/osdn/vnids.go
```diff
@@ -34,7 +69,7 @@ func (oc *OsdnController) GetVNID(name string) (uint, error) {
 		return id, nil
 	}
 	// In case of error, return some value which is not a valid VNID
-	return MaxVNID + 1, fmt.Errorf("Failed to find netid for namespace: %s in vnid map", name)
+	return netid.MaxVNID + 1, fmt.Errorf("Failed to find netid for namespace: %s in vnid map", name)
```
liggitt: returned uint shouldn't matter in an error case, use 0 rather than adding 1 to a number that could cause an overflow, depending on the constant value

plugins/osdn/vnids.go
```diff
@@ -55,7 +90,7 @@ func (oc *OsdnController) WaitAndGetVNID(name string) (uint, error) {
 	}
 
 	// In case of error, return some value which is not a valid VNID
-	return MaxVNID + 1, fmt.Errorf("Failed to find netid for namespace: %s in vnid map", name)
+	return netid.MaxVNID + 1, fmt.Errorf("Failed to find netid for namespace: %s in vnid map", name)
```
liggitt: same here, returned uint doesn't matter in an error case

plugins/osdn/vnids.go
```diff
@@ -9,22 +9,57 @@ import (
...
+		etcdAlloc = etcdallocator.NewEtcd(mem, "/ranges/namespacevnids", kapi.Resource("namespacevnidallocation"), oc.EtcdHelper)
+		return etcdAlloc
+	})
+
+	_, kclient := oc.Registry.GetSDNClients()
+
+	// Run vnid migration
+	migrate := netnamespace.NewMigrate(oc.EtcdHelper, kclient.Namespaces())
+	if err := migrate.Run(); err != nil {
+		return fmt.Errorf("Unable to migrate network namespaces: %v", err)
```
liggitt: what happens in this case? does the master fail to start up? what steps can they take to resolve the issue?

liggitt: also, does the migration remove the old NetNamespace objects? what happens if older masters/nodes are still paying attention to those?

pkg/netid/netid.go
pravisankar: No, this is not a new restriction. We have been using 10 as min vnid from the beginning. We don't expect namespace to have vnid between 1 to 9 but if that happens for some reason, vnid repair controller will detect this issue and won't allow starting master unless admin fixes this issue.

pkg/netid/netid.go
pravisankar: Ok, I will change the name to `pod.network.openshift.io/multitenant.requested-vnid`

plugins/osdn/vnids.go
pravisankar: Yes, returned uint shouldn't matter in case of error but to be on safer side, I'm trying to avoid the case where by mistake someone ignores the error and tries to use this value. 0 is valid vnid(admin network ID) and MaxVNID is  2^24, so we don't expect overflow and we have vnid range checks before persisting/assigning these ids.

pkg/cmd/admin/network/project_options.go
pravisankar: There will be error in this case but it won't tell this is due to old server and new client. I will throw useful error in this case.

plugins/osdn/vnids.go
pravisankar: Yes, master will not start if migration fails here. If the problem persists (very rare) they could delete NetNamespace objects manually using oc delete or curl on etcd and SDN will generate and assign new network IDs. This will lose prior config (project network merges/enabled global network access). Should I add this message as part of the error?
Yes, We remove old NetNamespace objects as part of migration. In case of HA master, SDN controller runs on one of the master. So there is no issue during rolling update. Migration order for master or node doesn't matter but all the nodes need to be migrated as well. Old nodes will watch for NetNamespace and this resource no longer exists. This info needs to be documented in the release notes?

plugins/osdn/vnids.go
liggitt: if someone is ignoring errors and using results anyway, nothing we can do will be safe.

plugins/osdn/vnids.go
liggitt: > If the problem persists (very rare) they could delete NetNamespace objects manually using oc delete or curl on etcd

they can't use `oc delete` if the master won't start, right? or does this kill controllers but not the master API?

> Migration order for master or node doesn't matter but all the nodes need to be migrated as well. Old nodes will watch for NetNamespace and this resource no longer exists. 

If I understand correctly, new nodes will only pay attention to namespace annotations, and old nodes will only pay attention to NetNamespace objects. As written, migration removes old objects as it creates the annotations, which means there is no upgrade order that ensures all actors have the information they need:
* upgrading nodes first means new nodes will be missing annotations on namespace until migration runs
* upgrading masters first means old nodes will miss NetNamespace objects until upgraded

I don't know what the nodes do with network setup when they are missing configured vnids, but it seems like it would be disruptive...

pravisankar: On Mon, Jun 6, 2016 at 10:39 AM, Jordan Liggitt <notifications@github.com>
wrote:

> In plugins/osdn/vnids.go
> <https://github.com/openshift/openshift-sdn/pull/303#discussion_r65933750>
> :
>
> > +		return fmt.Errorf("Unable to create NetID range: %v", err)
> > +	}
> > +
> > +	var etcdAlloc *etcdallocator.Etcd
> > +	netIDAllocator := vnidallocator.New(netIDRange, func(max int, rangeSpec string) allocator.Interface {
> > +		mem := allocator.NewContiguousAllocationMap(max, rangeSpec)
> > +		etcdAlloc = etcdallocator.NewEtcd(mem, "/ranges/namespacevnids", kapi.Resource("namespacevnidallocation"), oc.EtcdHelper)
> > +		return etcdAlloc
> > +	})
> > +
> > +	_, kclient := oc.Registry.GetSDNClients()
> > +
> > +	// Run vnid migration
> > +	migrate := netnamespace.NewMigrate(oc.EtcdHelper, kclient.Namespaces())
> > +	if err := migrate.Run(); err != nil {
> > +		return fmt.Errorf("Unable to migrate network namespaces: %v", err)
>
> If the problem persists (very rare) they could delete NetNamespace objects
> manually using oc delete or curl on etcd
>
> they can't use oc delete if the master won't start, right? or does this
> kill controllers but not the master API?
>
In case of HA setup, masters not running sdn controller can issue `oc
delete`. In case of single master, they can directly delete from etcd.


> Migration order for master or node doesn't matter but all the nodes need
> to be migrated as well. Old nodes will watch for NetNamespace and this
> resource no longer exists.
>
> If I understand correctly, new nodes will only pay attention to namespace
> annotations, and old nodes will only pay attention to NetNamespace objects.
> As written, migration removes old objects as it creates the annotations,
> which means there is no upgrade order that ensures all actors have the
> information they need:
>
>    - upgrading nodes first means new nodes will be missing annotations on
>    namespace until migration runs
>    - upgrading masters first means old nodes will miss NetNamespace
>    objects until upgraded
>
> I don't know what the nodes do with network setup when they are missing
> configured vnids, but it seems like it would be disruptive...
>
This won't be an issue. Event queues used in SDN watch routines repopulates
all the items every few minutes and if the item is not processed or failed
before will be re-attempted. So there will be some delay in syncing vnids
but migration order should not matter.


> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/openshift/openshift-sdn/pull/303/files/7918a79239fb027cd52a09a070fe2e9b40c0b7e4#r65933750>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABM0hkrz0AY7Zf9cUOzC1YCSNbbspkjTks5qJFtUgaJpZM4ITCfK>
> .
>



-- 
-Ravi

plugins/osdn/vnids.go
smarterclayton: We have a pretty strong convention that if you ignore an error we take you out back and break your fingers.  So anyone ignoring error (as jordan notes) deserves what they get.

pkg/netid/netid.go
```diff
@@ -0,0 +1,93 @@
...
+
+func SetVNID(ns *kapi.Namespace, id uint) error {
+	return setVNIDAnnotation(ns, VNIDAnnotation, id)
+}
+
+func DeleteVNID(ns *kapi.Namespace) {
+	delete(ns.Annotations, VNIDAnnotation)
+}
+
+func GetRequestedVNID(ns *kapi.Namespace) (uint, error) {
```
smarterclayton: This probably doesn't belong here - it belongs in `pkg/network/api/helper.go` or similar.

smarterclayton: But since you don't have that package it's fine, just add GoDoc.  We generally don't fiddle with this sort of helper wrapper.  Instead, we'd generally expect you to define a `type VNID uint` and have methods that return to and from it.

Again, fine for here.

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,232 @@
...
+type VnidController struct {
+	vnid   vnidallocator.Interface
+	client kclient.NamespaceInterface
+	Queue  *cache.EventQueue
+
+	// globalNamespaces are the namespaces that have access to all pods in the cluster and vice versa.
+	globalNamespaces sets.String
+}
+
+func NewVnidController(vnida vnidallocator.Interface, client kclient.NamespaceInterface, globalNamespaces []string) *VnidController {
```
smarterclayton: When you use VNID as an acronym in a name always use `VNID`.

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,232 @@
...
+	defer tx.Rollback()
+
+	var id uint
+	userRequestedVnid := false
+	value, err := netid.GetRequestedVNID(ns)
+	if err == nil {
+		id = value
+		userRequestedVnid = true
+		netid.DeleteRequestedVNID(ns)
+	} else if err != netid.ErrorVNIDNotFound {
```
smarterclayton: Handle errors first and return early, then handle normal flow.

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+			tx.Add(func() error { return c.vnid.Release(id) })
+		}
+	}
+	if err := netid.SetVNID(ns, id); err != nil {
+		return err
+	}
+
+	// TODO: could use a client.GuaranteedUpdate/Merge function
+	for i := 0; i < retryCount; i++ {
```
smarterclayton: We have `client.RetryOnConflict` to handle this use case.

plugins/osdn/netid/controller/controller.go
smarterclayton: If this line of debug is catching problems, we aren't doing enough stress testing of this flow in our cases.

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		glog.Errorf("Failed to apply vnid allocation, event-type: %v, namespace: %v, error: %v", eventType, ns, err)
+	}
+}
+
+// retryCount is the number of times to retry on a conflict when updating a namespace
+const retryCount = 2
+
+// addOrUpdate allocates network id(VNID) for the namespace if needed
+func (c *VnidController) addOrUpdate(ns *kapi.Namespace) error {
```
smarterclayton: Does this controller go away with Jordan's suggested changes for tolerating old and new?

plugins/osdn/netid/controller/controller.go
```diff
@@ -0,0 +1,232 @@
...
+
+	name := ns.ObjectMeta.Name
+	if userRequestedVnid {
+		// Only accept global VNID or VNIDs that are already allocated
+		if (id != netid.GlobalVNID) && !c.vnid.Has(id) {
+			return fmt.Errorf("requested netid %d not allocated", id)
+		}
+	} else {
+		// Do vnid allocation
+		if c.globalNamespaces.Has(name) {
```
smarterclayton: Why are you taking this as a config value when it could be stored on the namespace?

plugins/osdn/netid/controller/controller.go
smarterclayton: Ah, this is just cut and paste from the security controller.

plugins/osdn/netid/controller/controller.go
pravisankar: No, we still need this controller.

plugins/osdn/netid/controller/controller.go
pravisankar: This is not user config yet. c.globalNamespaces holds namespaces that we want to make it global by default (currently this is only done for 'default' namespace).

pravisankar:
@smarterclayton @liggitt updated, can you please take one more look?

pravisankar:
Closing this PR in favor of https://github.com/openshift/origin/pull/10365

PR #305, opened 2016-05-04
plugin reorg: merge subdirs, split master/node

danwinship:
The plugins/osdn vs plugins/osdn/ovs split was based on the idea that the code in plugins/osdn would eventually also be used by other plugins. Given that this is no longer expected to happen, there's no point in keeping the split.

OTOH, we *do* want to be able to split the master and node code apart more easily, and that's easier to do once everything is merged together, and that's what the second part of this branch does.

@openshift/networking PTAL. You'll probably want to review commit-by-commit.


plugins/osdn/plugin.go
```diff
@@ -44,6 +64,26 @@ func CreatePlugin(registry *Registry, multitenant bool, hostname string, selfIP
 	return plugin, err
 }
 
+func isOsdnPlugin(pluginType string) bool {
+	switch strings.ToLower(pluginType) {
+	case SingleTenantPluginName:
+		return true
+	case MultiTenantPluginName:
+		return true
```
pravisankar: L68-L72 can be simplified:
case SingleTenantPluginName, MultiTenantPluginName:
     return true

plugins/osdn/proxy.go
```diff
@@ -24,11 +28,17 @@ type ovsProxyPlugin struct {
...
-	}, nil
+// Called by higher layers to create the proxy plugin instance; only used by nodes
+func NewProxyPlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client) (api.FilteringEndpointsConfigHandler, error) {
+	switch strings.ToLower(pluginType) {
+	case MultiTenantPluginName:
+		return &ovsProxyPlugin{
+			registry: NewRegistry(osClient, kClient),
+			podsByIP: make(map[string]*kapi.Pod),
+		}, nil
+	}
```
pravisankar: If we expose isMultitenantPlugin() in plugins/osdn/plugin.go, then we can use it here:
```
if IsMultitenantPlugin(pluginType) {
   return &ovsProxyPlugin{...}
}
```

pravisankar:
We changed openshift-sdn-docker-setup.sh and openshift-sdn-ovs paths, need to update origin.spec and contrib/node/install-sdn.sh on the origin repo.

plugins/osdn/common.go
```diff
@@ -10,34 +10,28 @@ import (
...
-
-	AddServiceRules(service *kapi.Service, netID uint) error
-	DeleteServiceRules(service *kapi.Service) error
-
-	UpdatePod(namespace string, name string, id kubetypes.DockerID) error
-}
+const (
+	SingleTenantPluginName string = "redhat/openshift-ovs-subnet"
+	MultiTenantPluginName  string = "redhat/openshift-ovs-multitenant"
+)
```
pravisankar: Move these constants under pkg/ so that CLI and plugins/osdn can reuse it.

plugins/osdn/common.go
```diff
@@ -50,15 +44,29 @@ type OsdnController struct {
...
+	}
+	return createPlugin(osClient, kClient, isMultitenantPlugin(pluginType), "", "")
+}
+
+// Called by higher layers to create the plugin SDN node instance
+func NewNodePlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string) (api.OsdnPlugin, error) {
+	if !isOsdnPlugin(pluginType) {
+		return nil, nil
+	}
+	return createPlugin(osClient, kClient, isMultitenantPlugin(pluginType), hostname, selfIP)
```
pravisankar: We could pass pluginType to createPlugin() [L52, L60] and in L93, we can evaluate if the plugin is multitenant or not in one place:
```
plugin := &OsdnController{
 	multitenant:  IsMultitenantPlugin(pluginType) ,
...
```

plugins/osdn/node.go
```diff
@@ -0,0 +1,179 @@
...
+	mtu             uint
+}
+
+// Called by higher layers to create the plugin SDN node instance
+func NewNodePlugin(pluginType string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string, mtu uint) (api.OsdnNodePlugin, error) {
+	if !isOsdnPlugin(pluginType) {
+		return nil, nil
+	}
+
+	log.Infof("Starting with configured hostname '%s' (IP '%s')", hostname, selfIP)
```
pravisankar: include MTU in the log message?

danwinship:
Updated. I moved the constants into plugins/osdn/api/ rather than pkg/, since that seemed to be consistent with the rest of what we're doing.

Corresponding origin branch is "merge-and-split" in danwinship/origin

dcbw:
LGTM

pkg/cmd/admin/network/isolate_projects.go
```diff
@@ -9,6 +9,8 @@ import (
 	kcmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
 	kerrors "k8s.io/kubernetes/pkg/util/errors"
 
+	osdnapi "github.com/openshift/openshift-sdn/plugins/osdn/api"
```
pravisankar: I think we don't want pkg/cmd to depend on plugins/, please check https://github.com/openshift/openshift-sdn/pull/233

plugins/osdn/common.go
```diff
@@ -95,105 +30,6 @@ func isMultitenantPlugin(pluginType string) bool {
...
-		_, err := ipt.EnsureRule(iptables.Prepend, iptables.Table(rule.table), iptables.Chain(rule.chain), rule.args...)
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
-
 func GetNodeIP(node *kapi.Node) (string, error) {
```
pravisankar: We could move GetNodeIP() to plugins/osdn/subnets.go, that's where we handle node related stuff.

pravisankar: Also HostSubnetToString() definition in plugins/osdn/subnets.go ?

pkg/cmd/admin/network/isolate_projects.go
danwinship: OK then, for now I'm going to just revert the pkg/cmd/ part of the change; we already have origin/pkg/sdn/api and openshift-sdn/plugins/osdn/api, and the latter already depends on random stuff from kubernetes that we presumably don't want "oc" depending on, so it can't be moved wholesale to openshift-sdn/pkg/. So that means we would need some third "api" directory in openshift-sdn/pkg/ with some slightly different name, which will just be confusing.

We can revisit this in a later cleanup.

plugins/osdn/common.go
danwinship: Moved GetNodeIP(), left HostSubnetToString() since it's called from multiple files

danwinship:
Repushed with:

- pkg/cmd/ no longer imports plugin names from plugins/osdn/api
- common.go:GetNodeIP() is now subnets.go:getNodeIP()
- a bunch of other functions are also renamed to initial-lowercase since they don't need to be exported now that there's only a single namespace

I guess this should probably wait until after #303 though

pravisankar:
> Merge plugins/osdn/ovs into plugins/odsn
typo *osdn

pravisankar:
Good cleanup/reorg, LGTM

danwinship:
@openshift/networking updated; I ran the extended networking tests locally, but you might want to wait for the origin PR to finish the other tests before merging this

danwinship:
btw @pravisankar my version of this rebased on top of #303 is danwinship:merge-and-split-303, though that's based on an earlier version of this branch, and also I never tried compiling it, so it probably has bugs

danwinship:
dropped "Move plugin name constants into plugins/osdn/api/" since those constants are no longer used outside of openshift-sdn after Ravi's addition of IsOpenShiftNetworkPlugin() and IsOpenShiftMultiTenantNetworkPlugin().

dcbw:
LGTM

plugins/osdn/common.go
```diff
@@ -10,33 +10,22 @@ import (
...
-	AddServiceRules(service *kapi.Service, netID uint) error
-	DeleteServiceRules(service *kapi.Service) error
-
-	UpdatePod(namespace string, name string, id kubetypes.DockerID) error
-}
-
 type OsdnController struct {
-	pluginHooks        PluginHooks
 	pluginName         string
+	multitenant        bool
```
pravisankar: We don't need `multitenant` anymore?
Now we got the pluginName, can use IsOpenShiftMultitenantNetworkPlugin(oc.pluginName) where ever required.

plugins/osdn/node.go
```diff
@@ -0,0 +1,140 @@
...
+	localSubnet        *osapi.HostSubnet
+	hostName           string
+	podNetworkReady    chan struct{}
+	vnids              vnidMap
+	iptablesSyncPeriod time.Duration
+	mtu                uint
+}
+
+// Called by higher layers to create the plugin SDN node instance
+func NewNodePlugin(pluginName string, osClient *osclient.Client, kClient *kclient.Client, hostname string, selfIP string, iptablesSyncPeriod time.Duration, mtu uint) (api.OsdnNodePlugin, error) {
```
pravisankar: We have NewNodePlugin(), NewProxyPlugin() but for master it is StartMaster()..can we reintroduce NewMasterPlugin() to be consistent?

plugins/osdn/node.go
```diff
@@ -0,0 +1,140 @@
...
+
+	kapi "k8s.io/kubernetes/pkg/api"
+	kclient "k8s.io/kubernetes/pkg/client/unversioned"
+	kubeletTypes "k8s.io/kubernetes/pkg/kubelet/container"
+	kexec "k8s.io/kubernetes/pkg/util/exec"
+	kubeutilnet "k8s.io/kubernetes/pkg/util/net"
+)
+
+type OsdnNode struct {
+	multitenant        bool
```
pravisankar: Store pluginName instead, we can always deduce if it is multitenant or not.

plugins/osdn/common.go
danwinship: Later in the branch, both pluginName and multitenant go away in OsdnMaster, and pluginName goes away in OsdnNode. (Keeping oc.multitenant rather than keeping oc.pluginName and using IsOpenShiftMultiTenantNetworkPlugin() involved less code change.)

plugins/osdn/node.go
danwinship: We could... but look at the code in origin: https://github.com/openshift/origin/pull/9476/files#diff-232120b6845c4d46e68c543a21dbf8c8L377. The master plugin object didn't get used anyway, origin just created it, called Start() on it, and then forgot about it.

I think with a bit of further rewriting we should eventually be able to similarly squash the node and proxy code too; right now they're created and started in different files, but I don't think that's actually necessary.

plugins/osdn/node.go
danwinship: I could, but after creation time, we'd never actually use the plugin name for anything other than checking if it's multitenant, so that kind of seems like a waste. (Well, ok, plugin.Name() would make more sense if we had the pluginName rather than multitenant, I guess...)

So should I change it or not?

pravisankar: I thought we could use pluginName for logging instead of saying single-tenant/multi-tenant plugin but looks like we are already logging pluginName as part of NewNodePlugin(). I'm fine with the current code. Since this is internal stuff we can store pluginName later when needed.

plugins/osdn/node.go
pravisankar: ok

pravisankar:
LGTM

PR #306, opened 2016-05-10
Add tc state to the debug tar

knobunc:
Added the output from tc sub-commands to the output so we can see the traffic shaper state.


dcbw:
LGTM

PR #308, opened 2016-05-12
Remove prompts from commented examples

rhcarvalho:
Following the upstream style changes from Kubernetes and OpenShift
Origin.

See https://github.com/openshift/origin/pull/8454.


rhcarvalho:
@openshift/networking PTAL

danwinship:
OK

PR #309, opened 2016-05-12
debug: log master config and master api/controllers journal and service

dcbw:


danwinship:
See discussion on #287 by @knobunc. This has the same problem, that it assumes there will be a config file for the master, but some deployments (including, I believe, our default non-ha ansible deployment) don't have any such file; they just use command-line arguments.

hack/debug.sh
```diff
@@ -73,7 +90,13 @@ do_master () {
     mkdir -p $logmaster
 
     # Log the generic system stuff
-    log_system $logmaster $aos_master_service
+    services=
+    for service in "${services[@]}"; do
+        if [ -n "${service}" ]; then
+            services="${services} ${service}"
+        fi
```
danwinship: You're overwriting the earlier "services" variable (`services=($aos_master_service $aos_master_controllers_service $aos_master_api_service)`) here so this loop is going to just fail isn't it? If it does actually work due to perl-like separation of scalar and array variables, it's at least incredibly confusing...

dcbw:
@danwinship good point; updated.  HA stuff *must* use a config file at this point, since the master controller service needs it.  But non-HA definitely can just use command-line args.

hack/debug.sh
```diff
@@ -61,6 +68,15 @@ log_system () {
 }
 
 do_master () {
+    services=($aos_master_service $aos_master_controllers_service $aos_master_api_service)
+    start_args=("master" "master controllers" "master api")
+    for i in "${!services[@]}"; do
+        config=$(get_config_path_from_service "${start_args[$i]}" "${services[$i]}")
+        if [ -n "$config" ]; then
+            break
```
pravisankar: What if 'master controller' and 'master api' services are running on the same host? This will only pick one config.

pravisankar: Both the services should use the same master config, so it shouldn't be an issue.

hack/debug.sh
```diff
@@ -73,7 +89,13 @@ do_master () {
     mkdir -p $logmaster
 
     # Log the generic system stuff
-    log_system $logmaster $aos_master_service
+    log_services=
+    for service in "${services[@]}"; do
+        if [ -n "${service}" ]; then
+            log_services="${log_services} ${service}"
+        fi
+    done
```
pravisankar: L92-L97:
log_services=${services[*]} ?

danwinship: I guess I didn't look closely at this code before... what it is even doing? ${services} should contain exactly three non-empty strings...

dcbw: My bash is not the best, as you've found.  ${services[*]} would certainly be easier.

dcbw:
@danwinship @pravisankar PTAL; now each config file (if it exists) is added to the tarball along with its service.  The 'zzzz' seems necessary since when passing an array through to a bash function the array gets expanded and empty elements won't be present, so otherwise the indexes would no longer be the same between service_names and config_files.  But that may also just be my bash-inexperience.  Still, it produces the right result for me...

hack/debug.sh
```diff
@@ -61,6 +70,16 @@ log_system () {
 }
 
 do_master () {
+    services=($aos_master_service $aos_master_controllers_service $aos_master_api_service)
+    start_args=("master" "master controllers" "master api")
+    config_files=()
+    for i in "${!services[@]}"; do
+        config_files[$i]=$(get_config_path_from_service "${start_args[$i]}" "${services[$i]}")
+        if [ -z "${config_files[$i]}" ]; then
+            config_files[$i]="zzzz"
```
danwinship: maybe it would be easier to split a `log_service` function out of log_system and then just call that once for each service (below, after calling `log_system`) rather than trying to deal with bash arrays

hack/debug.sh
```diff
@@ -253,11 +272,26 @@ do_pod_service_connectivity_check () {
...
+    # 'atomic-openshift-node.service', 'atomic-openshift-master-controllers.service', etc
+    service_name=$2
+    config=$(ps wwaux | grep -v grep | sed -ne "s/.*openshift start ${service_type}.*--config=\([^ ]*\.yaml\).*/\1/p")
     if [ -z "$config" ]; then
-	config=$(systemctl show -p ExecStart $aos_node_service | sed -ne 's/.*--config=\([^ ]*\).*/\1/p')
+	config=$(systemctl show -p ExecStart $service_name | sed -ne 's/.*--config=\([^ ]*\).*/\1/p')
+	if [ "$config" == "\${CONFIG_FILE}" ]; then
+	    varfile=$(systemctl show $service_name | grep EnvironmentFile | sed -ne 's/EnvironmentFile=\([^ ]*\).*/\1/p')
+	    if [ -f "$varfile" ]; then
+	        config=$(cat $varfile | grep CONFIG_FILE | sed -ne 's/CONFIG_FILE=//p')
```
danwinship: the `grep` is redundant with the `sed -n`

hack/debug.sh
```diff
@@ -468,6 +502,8 @@ for systemd_dir in /etc/systemd/system /usr/lib/systemd/system; do
     for name in openshift origin atomic-openshift; do
 	if [ -f $systemd_dir/$name-master.service ]; then
 	    aos_master_service=$name-master.service
+	    aos_master_controllers_service=$name-master-controllers.service
+	    aos_master_api_service=$name-master-api.service
```
danwinship: Will `atomic-openshift-master.service` always exist on HA systems, or should we do `if [ -f ...` tests for each file separately?

hack/debug.sh
```diff
@@ -253,11 +272,26 @@ do_pod_service_connectivity_check () {
...
 }
 
-do_node () {
-    config=$(ps wwaux | grep -v grep | sed -ne 's/.*openshift.*--config=\([^ ]*\.yaml\).*/\1/p')
+get_config_path_from_service() {
+    # 'node', 'master', 'master api', 'master controllers'
+    service_type=$1
+    # 'atomic-openshift-node.service', 'atomic-openshift-master-controllers.service', etc
+    service_name=$2
+    config=$(ps wwaux | grep -v grep | sed -ne "s/.*openshift start ${service_type}.*--config=\([^ ]*\.yaml\).*/\1/p")
```
danwinship: if `${service_type}` is "master", then this will match "openshift start master api" and "openshift start master controllers" too, which we don't want, right?

hack/debug.sh
pravisankar: You could use associative sub-arrays in bash:
service_config_files = ( (service1, config1) (service2, config2)...)
But I prefer what @danwinship suggested, simpler and don't have to deal with bash arrays.

hack/debug.sh
pravisankar: Are these service names going to change due to new branding? (openshift local/lab/enterprise)

dcbw:
@danwinship @pravisankar @knobunc PTAL, thanks.

Yes, the branding/services may change, but we'll have to update the script for that anyway...

danwinship:
assuming it's tested, LGTM

dcbw:
@danwinship I ran it on an HA cluster and it spits out the right logs.  I haven't run it on a non-HA cluster though, but I'll do that today.

dcbw:
FWIW, the log output now looks like this...

> ./master/journal-atomic-openshift-master.service
> ./master/systemctl-show-atomic-openshift-master.service
> ./master/journal-atomic-openshift-master-controllers.service
> ./master/systemctl-show-atomic-openshift-master-controllers.service
> ./master/CONFIG-atomic-openshift-master-controllers.service
> ./master/journal-atomic-openshift-master-api.service
> ./master/systemctl-show-atomic-openshift-master-api.service
> ./master/CONFIG-atomic-openshift-master-api.service

and for a non-HA master:

> openshift-sdn-debug/master/journal-openshift-master.service
> openshift-sdn-debug/master/systemctl-show-openshift-master.service

hack/debug.sh
```diff
@@ -253,24 +275,42 @@ do_pod_service_connectivity_check () {
...
 }
 
-do_node () {
-    config=$(ps wwaux | grep -v grep | sed -ne 's/.*openshift.*--config=\([^ ]*\.yaml\).*/\1/p')
+get_config_path_from_service() {
+    # 'node', 'master', 'master api', 'master controllers'
+    service_type=$1
+    # 'atomic-openshift-node.service', 'atomic-openshift-master-controllers.service', etc
+    service_name=$2
+    config=$(ps wwaux | grep -v grep | sed -ne "s/.*openshift start ${service_type} --config=\([^ ]*\.yaml\).*/\1/p")
```
pravisankar: This will not match in some cases where the service uses other cli options before --config.
Example: /usr/bin/openshift start node --loglevel=5 --config=/...
ps wwaux | grep -v grep | sed -ne "s/.*openshift start node --.* --config=\([^ ]*\.yaml\).*/\1/p" ?

Issue #307, opened 2016-05-11
Host IP changes should update subnet instead of delete+create subnet


pravisankar:
In OpenShift SDN, all the nodes watch for subnet changes and updates their hostsubnet rules. Any host IP change will appear as node unregister and register of new node. Supporting update subnet will avoid this confusion and it will also reduce calls to etcd store and watch processes.

PR #310, opened 2016-05-13
Update subnet when host IP changes instead of delete-old + create-new subnet

pravisankar:
closes: https://github.com/openshift/openshift-sdn/issues/307

### DIFF
--- a/plugins/osdn/ovs/proxy.go
+++ b/plugins/osdn/ovs/proxy.go
@@ -3,6 +3,7 @@ package ovs
 import (
 	"fmt"
 	"net"
+	"strings"
 	"sync"
 
 	"github.com/golang/glog"
@@ -63,6 +64,7 @@ func (proxy *ovsProxyPlugin) watchPods() {
 		}
 		pod := obj.(*kapi.Pod)
 
+		glog.V(5).Infof("Watch %s event for Pod %q", strings.Title(string(eventType)), pod.ObjectMeta.Name)
 		switch eventType {
 		case watch.Added, watch.Modified:
 			proxy.trackPod(pod)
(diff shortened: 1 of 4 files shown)


pravisankar:
Origin PR: https://github.com/openshift/origin/pull/8879

pravisankar:
@openshift/networking PTAL

plugins/osdn/ovs/proxy.go
```diff
@@ -65,8 +65,12 @@ func (proxy *ovsProxyPlugin) watchPods() {
 
 		switch eventType {
 		case watch.Added, watch.Modified:
+			glog.V(5).Infof("Watch Added or Modified event for Pod %q", pod.ObjectMeta.Name)
```
danwinship: any reason to not do `"Watch %s event...", string(eventType)`? (Or even `strings.Title(string(eventType))`)

danwinship:
LGTM

plugins/osdn/ovs/proxy.go
pravisankar: +1, I'll update

plugins/osdn/subnets.go
```diff
@@ -50,52 +50,40 @@ func (oc *OsdnController) addNode(nodeName string, nodeIP string) error {
...
 	if err != nil {
-		// Release subnet if needed
-		if subnetAllocated {
-			_, ipnet, er := net.ParseCIDR(subnetCIDR)
-			if er != nil {
-				log.Errorf("Failed to release subnet %q for node %q: %v", subnetCIDR, nodeName, er)
-			} else {
-				oc.subnetAllocator.ReleaseNetwork(ipnet)
-			}
+		_, ipnet, er := net.ParseCIDR(subnetCIDR)
```
dcbw: Couldn't we just use "sn" here?

dcbw:
Small nit, otherwise LGTM

pravisankar:
@dcbw @danwinship  Updated

danwinship:
LGTM but lets get the origin side in first since this depends on it

PR #313, opened 2016-05-24
Periodically sync openshift node iptables rules

pravisankar:


pravisankar:
@openshift/networking PTAL

plugins/osdn/node_firewall.go
```diff
@@ -0,0 +1,102 @@
...
+	utilwait "k8s.io/kubernetes/pkg/util/wait"
+)
+
+type FirewallRule struct {
+	table string
+	chain string
+	args  []string
+}
+
+type Firewall struct {
```
danwinship: "firewall" isn't the right name (for the struct or the file); half of the iptables rules are to *un*-firewall traffic, and the other half are NAT rules that are needed regardless of firewalling. So the name should be something with "iptables", not "firewall"

plugins/osdn/node_firewall.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return nil
+}
+
+// Get openshift iptables rules
+func (f *Firewall) getStaticNodeFirewallRules() []FirewallRule {
+	return []FirewallRule{
+		{"nat", "POSTROUTING", []string{"-s", f.clusterNetworkCIDR, "!", "-d", f.clusterNetworkCIDR, "-j", "MASQUERADE"}},
+		{"filter", "INPUT", []string{"-p", "udp", "-m", "multiport", "--dports", "4789", "-m", "comment", "--comment", "001 vxlan incoming", "-j", "ACCEPT"}},
+		{"filter", "INPUT", []string{"-i", "tun0", "-m", "comment", "--comment", "traffic from docker for internet", "-j", "ACCEPT"}},
```
danwinship: Can you arrange for `ovs.TUN` to get passed down to here so that we don't have to hardcode `"tun0"`? (If it's not easy/clean then never mind, it will probably be easier after #305.)

danwinship:
Note gofmt errors in travis.

Basically LGTM. No substantive objections.

pravisankar:
@danwinship Updated, PTAL

dcbw:
LGTM

PR #314, opened 2016-05-26
Fail more cleanly on script errors

danwinship:
If we fail to get info from docker/ip/etc, fail immediately rather than failing with a syntax error due to substituting in an empty value later.

inspired by https://bugzilla.redhat.com/show_bug.cgi?id=1333393, though as noted there, we probably want a better solution eventually to ensure we don't leave cruft in OVS

@openshift/networking PTAL


pravisankar:
LGTM

dcbw:
LGTM

PR #315, opened 2016-05-31
Force a rebuild after ./hack/sync-to-origin.sh

danwinship:
I don't know if it's go's fault or godep or our Makefiles or what, but something isn't very good about noticing that Godeps/_workspace/src/github.com/openshift/openshift-sdn/ has changed, so I tend to have to "make clean; make" after doing a sync-to-origin. This patch "fixes" things by deleting the openshift-sdn .a files after a sync so that the next origin make will have to rebuild them. (I thought about trying to make this a bit more clever about which .a files it deletes, but openshift-sdn is small enough that it doesn't take that much time to rebuild all of it anyway...)

@openshift/networking PTAL

### DIFF
--- a/hack/sync-to-origin.sh
+++ b/hack/sync-to-origin.sh
@@ -35,13 +35,13 @@ validate_args() {
     sdn_repo_dir=$(readlink -f -- $sdn_repo_dir)
 }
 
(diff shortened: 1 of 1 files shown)


pravisankar:
LGTM

PR #316, opened 2016-05-31
debug.sh: Fix fetching config file for service

pravisankar:
This will also handle the case where optional config options are used before --config
Example: /usr/bin/openshift start node --loglevel=5 --config=/...


pravisankar:
@dcbw @danwinship PTAL

danwinship:
LGTM

dcbw:
Ah yeah, that's a better fix than what I was thinking.  Thanks!

PR #317, opened 2016-05-31
Simplify caching and fetching network information

pravisankar:
Precursor for storing/fetching plugin name from ClusterNetwork.

### DIFF
--- a/plugins/osdn/common.go
+++ b/plugins/osdn/common.go
@@ -92,7 +92,7 @@ func (oc *OsdnController) BaseInit(registry *Registry, pluginHooks PluginHooks, 
 	return nil
 }
 
-func (oc *OsdnController) validateNetworkConfig(clusterNetwork, serviceNetwork *net.IPNet) error {
+func (oc *OsdnController) validateNetworkConfig(ni *NetworkInfo) error {
 	// TODO: Instead of hardcoding 'tun0' and 'lbr0', get it from common place.
 	// This will ensure both the kube/multitenant scripts and master validations use the same name.
 	hostIPNets, err := netutils.GetHostIPNetworks([]string{"tun0", "lbr0"})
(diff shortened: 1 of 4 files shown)


pravisankar:
@openshift/networking PTAL

danwinship:
LGTM

plugins/osdn/common.go
```diff
@@ -140,64 +140,64 @@ func (oc *OsdnController) validateNetworkConfig(clusterNetwork, serviceNetwork *
...
 		return true, nil
 	}
 	return false, nil
 }
 
 func (oc *OsdnController) StartMaster(clusterNetworkCIDR string, clusterBitsPerSubnet uint, serviceNetworkCIDR string) error {
 	// Validate command-line/config parameters
-	hostBitsPerSubnet := int(clusterBitsPerSubnet)
-	clusterNetwork, _, serviceNetwork, err := ValidateClusterNetwork(clusterNetworkCIDR, hostBitsPerSubnet, serviceNetworkCIDR)
+	network, err := ValidateClusterNetwork(clusterNetworkCIDR, int(clusterBitsPerSubnet), serviceNetworkCIDR)
```
dcbw: Maybe 'network' should be 'ni' here to be consistent with other parts of the code?  I won't block a merge for that though.

plugins/osdn/registry.go
```diff
@@ -126,75 +132,63 @@ func (registry *Registry) GetPod(nodeName, namespace, podName string) (*kapi.Pod
...
-	}
-	return registry.clusterNetwork, registry.hostSubnetLength, registry.serviceNetwork, nil
-}
-
-func (registry *Registry) GetClusterNetwork() (*net.IPNet, error) {
-	if err := registry.cacheClusterNetwork(); err != nil {
 		return nil, err
 	}
-	return registry.clusterNetwork, nil
+	return ValidateClusterNetwork(cn.Network, cn.HostSubnetLength, cn.ServiceNetwork)
```
dcbw: Shouldn't this cache the return value of ValidateClusterNetwork() into registry.NetworkInfo?  I can't find where the cached object gets set, but I'm probably missing it while skimming the patch.

pravisankar:
@dcbw @danwinship Updated, PTAL

dcbw:
LGTM now

PR #318, opened 2016-06-02
Persist network plugin name as part of cluster network creation

pravisankar:
Depends on https://github.com/openshift/openshift-sdn/pull/317

Trello card: https://trello.com/c/Lr0wa4hr/228-3-detect-master-and-node-sdn-plugin-mismatch-sdn

### DIFF
--- a/plugins/osdn/common.go
+++ b/plugins/osdn/common.go
@@ -252,5 +252,5 @@ func GetPodContainerID(pod *kapi.Pod) string {
 }
 
 func HostSubnetToString(subnet *osapi.HostSubnet) string {
-	return fmt.Sprintf("%s [host: '%s'] [ip: '%s'] [subnet: '%s']", subnet.Name, subnet.Host, subnet.HostIP, subnet.Subnet)
+	return fmt.Sprintf("%s (host: %q, ip: %q, subnet: %q)", subnet.Name, subnet.Host, subnet.HostIP, subnet.Subnet)
 }
--- a/plugins/osdn/common.go
+++ b/plugins/osdn/common.go
@@ -36,6 +36,7 @@ type PluginHooks interface {
 
 type OsdnController struct {
 	pluginHooks        PluginHooks
+	pluginName         string
 	Registry           *Registry
 	localIP            string
 	localSubnet        *osapi.HostSubnet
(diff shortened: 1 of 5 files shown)


pravisankar:
@openshift/networking PTAL

danwinship:
LGTM

plugins/osdn/common.go
```diff
@@ -140,64 +138,72 @@ func (oc *OsdnController) validateNetworkConfig(clusterNetwork, serviceNetwork *
...
-		log.Errorf("Failed to obtain ClusterNetwork: %v", err)
-		return err
+		return fmt.Errorf("Failed to get network information: %v", err)
+	}
+	if ni.PluginName != oc.pluginName {
+		if ni.PluginName == "" {
+			// Do not return error in this case
+			log.Warningf("Either there is network plugin mismatch between OpenShift master(%q) and node(%q) or OpenShift master is running older version where we don't persist plugin name")
+		} else {
+			return fmt.Errorf("Detected network plugin mismatch between OpenShift master(%q) and node(%q)", ni.PluginName, oc.pluginName)
```
danwinship: should we (in another PR) deprecate the node-config networkPluginName argument and just get that value from the master like we do with the rest of the network config?

plugins/osdn/ovs/plugin.go
```diff
@@ -35,10 +35,18 @@ type ovsPlugin struct {
 	multitenant bool
 }
 
-func CreatePlugin(registry *osdn.Registry, multitenant bool, hostname string, selfIP string, iptablesSyncPeriod time.Duration) (api.OsdnPlugin, error) {
-	plugin := &ovsPlugin{multitenant: multitenant}
+func isPluginMultitenant(pluginName string) bool {
+	switch strings.ToLower(pluginName) {
```
dcbw: how about just "return pluginName == MultiTenantPluginName" ?

dcbw:
Needs a rebase onto the updated ClusterNetwork caching cleanup, but overall looks OK to me.

pravisankar:
@openshift/networking updated, PTAL

dcbw:
LGTM

PR #320, opened 2016-06-10
[DO NOT MERGE] Convert OpenShift SDN to a CNI plugin

dcbw:
Based off @danwinship's plugin reorg branch, so start with the "Use CNI for IPAM instead of docker" commit.  Not quite ready for merge yet, but I'd love feedback on the approach, especially the pod info server parts.

cc @pravisankar @rajatchopra @smarterclayton


plugins/osdn/bin/openshift-sdn-docker-setup.sh
```diff
@@ -1,26 +0,0 @@
-#!/bin/bash
-
-set -ex
-
-bridge=$1
-mtu=$2
-
-DOCKER_NETWORK_OPTIONS="-b=${bridge} --mtu=${mtu}"
```
danwinship: This causes docker to set the correct MTU on the veths. You need to add that functionality somewhere else if docker isn't going to do it any more.

plugins/osdn/controller.go
```diff
@@ -214,10 +126,6 @@ func (plugin *OsdnNode) SetupSDN(localSubnetCIDR, clusterNetworkCIDR, servicesNe
 	otx.AddFlow("table=0, priority=200, in_port=2, arp, nw_src=%s, nw_dst=%s, actions=goto_table:5", localSubnetGateway, clusterNetworkCIDR)
 	otx.AddFlow("table=0, priority=200, in_port=2, ip, actions=goto_table:5")
 	otx.AddFlow("table=0, priority=150, in_port=2, actions=drop")
-	// vovsbr
-	otx.AddFlow("table=0, priority=200, in_port=3, arp, nw_src=%s, actions=goto_table:5", localSubnetCIDR)
-	otx.AddFlow("table=0, priority=200, in_port=3, ip, nw_src=%s, actions=goto_table:5", localSubnetCIDR)
-	otx.AddFlow("table=0, priority=150, in_port=3, actions=drop")
```
danwinship: You need to change the `actions=output:3` rules in table 6 and table 7 to `actions=drop` as well

plugins/osdn/plugin.go
```diff
@@ -173,8 +180,23 @@ func (plugin *OsdnNode) SetUpPod(namespace string, name string, id kubeletTypes.
...
+		return err
+	}
+
+	containerNs, err := plugin.getContainerNS(id)
+	if err != nil {
+		return fmt.Errorf("Failed to get container network namespace: %v", err)
+	}
+	defer containerNs.Close()
+
+	if _, err := plugin.runIPAM(containerNs, "DEL", id); err != nil {
```
danwinship: it's weird that TearDownPod() calls plugin.runIPAM("DEL") but SetUpPod() doesn't call plugin.runIPAM("ADD"). (because in one case it's hidden in a helper function and in the other case it isn't...)

plugins/osdn/plugin.go
```diff
@@ -173,8 +180,23 @@ func (plugin *OsdnNode) SetUpPod(namespace string, name string, id kubeletTypes.
...
+		return err
+	}
+
 	// The script's teardown functionality doesn't need the VNID
-	out, err := exec.Command(plugin.getExecutable(), tearDownCmd, string(id), "-1", "-1", "-1").CombinedOutput()
+	out, err := exec.Command(plugin.getExecutable(), tearDownCmd, hostVethName, contMac, contIP, "-1", "", "", "false").CombinedOutput()
```
danwinship: note that in addition to the dummy arguments that are already ""/"-1", hostVethName and contMac are also unused here. Maybe we should just make setup and teardown take different args? (But maybe this will all eventually go away and it doesn't matter.)

plugins/osdn/plugin.go
```diff
@@ -206,3 +243,218 @@ func (plugin *OsdnNode) UpdatePod(namespace string, name string, id kubeletTypes
...
+	}
+
+	// Clear docker addresses and routes from container veth
+	err = containerNs.Do(func (ns.NetNS) error {
+		addrs, err := netlink.AddrList(contVeth, syscall.AF_UNSPEC)
+		if err != nil {
+			return err
+		}
+		for _, a := range addrs {
+			netlink.AddrDel(contVeth, &a)
```
danwinship: grumble. should either use pkg/ipcmd or else port the existing stuff too...

plugins/osdn/plugin.go
```diff
@@ -206,3 +243,218 @@ func (plugin *OsdnNode) UpdatePod(namespace string, name string, id kubeletTypes
...
+		Command:     action,
+		ContainerID: string(id),
+		NetNS:       containerNs.Path(),
+		IfName:      containerIfname,
+		Path:        "/opt/cni/bin",
+	}
+
+	// run the IPAM plugin and get back the config to apply
+	var podIP string
+	if action == "ADD" {
```
danwinship: kind of bad go style to have a string literal rather than a constant

plugins/osdn/plugin.go
```diff
@@ -206,3 +243,218 @@ func (plugin *OsdnNode) UpdatePod(namespace string, name string, id kubeletTypes
...
+		}
+
+		err = containerNs.Do(func(ns.NetNS) error {
+			return ipam.ConfigureIface(containerIfname, result)
+		})
+		if err != nil {
+			return "", fmt.Errorf("Failed to configure container interface: %v", err)
+		}
+
+		podIP = result.IP4.IP.IP.String()
```
danwinship: Maybe say "IP" a few more times, just to be safe :)

plugins/osdn/master.go
```diff
@@ -18,7 +18,7 @@ import (
 type OsdnMaster struct {
 	registry        *Registry
 	subnetAllocator *netutils.SubnetAllocator
-	vnidMap         VNIDMap
+	vnidMap         *VNIDMap
```
danwinship: should probably be a separate commit

plugins/osdn/podinfoserver.go
```diff
@@ -0,0 +1,138 @@
...
+	"path"
+	"syscall"
+
+	"github.com/gorilla/mux"
+
+	"github.com/openshift/openshift-sdn/pkg/netutils"
+	"github.com/openshift/openshift-sdn/plugins/osdn/api"
+)
+
+type PodInfoServer struct {
```
danwinship: it's not just a *pod* info server though; it serves ClusterNetwork too...

plugins/osdn/podinfoserver.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	dirName := path.Dir(api.PodInfoSocketPath)
+	if err := os.MkdirAll(dirName, 0700); err != nil {
+		return fmt.Errorf("failed to create pod info socket directory: %v", err)
+	}
+
+	mask := syscall.Umask(0777)
+	l, err := net.Listen("unix", api.PodInfoSocketPath)
+	syscall.Umask(mask)
```
danwinship: That is not thread-safe...
Wait, and then you chmod it to 0600 afterward anyway??

plugins/osdn/api/types.go
```diff
@@ -1,13 +1,10 @@
...
 import (
-	knetwork "k8s.io/kubernetes/pkg/kubelet/network"
 	pconfig "k8s.io/kubernetes/pkg/proxy/config"
 )
 
 type OsdnNodePlugin interface {
-	knetwork.NetworkPlugin
-
 	Start() error
 }
```
danwinship: can probably just get rid of OsdnNodePlugin in this case like I did with OsdnMasterPlugin?

plugins/osdn/node.go
```diff
@@ -139,6 +148,46 @@ func (node *OsdnNode) Start() error {
...
+const cniConfigPath = "/etc/cni/net.d/80-openshift-sdn.conf"
+
+func writeCniConfig() error {
+	const config string =`{
+  "cniVersion": "0.1.0",
+  "name": "openshift-sdn",
+  "type": "openshift-sdn"
+}`
+
+	return ioutil.WriteFile(cniConfigPath, []byte(config), 0600)
```
danwinship: We should not be writing files to /etc at runtime. That should be installed with the plugin.

plugins/osdn/sdn-cni-plugin/openshift-sdn.go
```diff
@@ -0,0 +1,460 @@
...
+}`
+)
+
+func getNetworkInfo(client *http.Client) (*api.GetNetworkResponse, error) {
+	// Try few times up to 15 seconds
+	retries := 150
+	retryInterval := 100 * time.Millisecond
+	for i := 0; i < retries; i++ {
+		network := &api.GetNetworkResponse{}
+		if err := getUrl(client, "http://d/network", network); err == nil {
```
danwinship: "d"?

plugins/osdn/podinfoserver.go
dcbw: Fair enough; originally it was just the pod info stuff and I was going to write the cluster info to /etc/cni/net.d/openshift-sdn.conf.  But it turns out kubelet's CNI plugin needs that file immediately at startup, and we don't have cluster info until later so I couldn't do that.  I'll fix this up.

plugins/osdn/node.go
dcbw: As commented above, originally I was going to write the cluster info to this file as well, but kubelet's CNI plugin needs this file immediately, and we don't have cluster info til later.  So you're right, this can now be dropped statically on the filesystem.

plugins/osdn/podinfoserver.go
dcbw: Does it need to be thread-safe?  It's just creating the pod info server, which only happens once at node startup.  It won't ever run from two threads at the same time.

danwinship: I meant that you might be changing the umask for an unrelated operation happening at the same time in another thread.

plugins/osdn/api/types.go
dcbw: It's more complicated for the node, because the OSDN plugin gets constructed (node_config.go::BuildKubernetesNodeConfig()) in a different place than it gets started (node.go::RunPlugin()).

plugins/osdn/controller.go
dcbw: Fixed.

dcbw:
@danwinship PTAL, thanks!

plugins/osdn/podinfoserver.go
```diff
@@ -0,0 +1,138 @@
...
+		registry: r,
+		vnidMap:  v,
+		nodeName: nodeName,
+		Server:   http.Server{
+			Handler: router,
+		},
+	}
+
+	router.NotFoundHandler = http.HandlerFunc(http.NotFound)
+	router.Methods("GET").Path("/network").HandlerFunc(s.getNetwork)
```
smarterclayton: I am *really* uncomfortable about creating a new API here (which is what this is) and also not happy to see a new set of API definitions that don't overlap with our existing objects being created.  I need to dive into this more and understand the tradeoffs, but like to see a quick writeup of what options you considered and why you chose this approach.

One question I was surprised by - why does the CNI interface not pass enough information down to the network plugin to make a decision there?

plugins/osdn/sdn-cni-plugin/openshift-sdn.go
```diff
@@ -0,0 +1,462 @@
...
+
+	return info, nil
+}
+
+// TODO: replace with k8s.io/kubernetes/pkg/util/bandwidth when rebasing to Kube 1.3
+var minRsrc = resource.MustParse("1k")
+var maxRsrc = resource.MustParse("1P")
+
+func validateBandwidthIsReasonable(rsrc *resource.Quantity) error {
+	if rsrc.Value() < minRsrc.Value() {
```
smarterclayton: This doesn't take into account rollover.  You have to use quantity.Cmp()

smarterclayton:
Do you guys have a design doc / writeup / proposal to cover this work?

plugins/osdn/sdn-cni-plugin/openshift-sdn.go
dcbw: It's basically C&P from upstream kubernetes code, and we'll replace it with that kubernetes code when the 1.3 rebase happens.  So does this mean the upstream code is also wrong?

plugins/osdn/podinfoserver.go
dcbw: @smarterclayton I would not consider the podinfoserver new API; it's private IPC between the openshift-sdn CNI plugin, and the openshift-sdn node plugin.  It is not intended to be public in any way, and we would reserve the right to change it at any time.

The options:

* root-only private unix domain socket: advantages include a very small amount of well-understood code doing very specific things, which closely follows other examples like Docker's API server and client pattern.  Authentication and security are protected by simple, common Unix mechanisms like filesystem permissions, and we could additionally protect it through peer credentials if we wanted to.  The private API exposed by this socket is small and easily extendable if/when the SDN plugin needs to extend it, but we do not expect or support any other consumers of this API than openshift-sdn.
* talk to api-server: this would require a larger amount of code in the CNI plugin to create Kube and OpenShift clients and find/read the Pod, VNID, HostSubnet, and other objects the CNI plugin requires.  The openshift-sdn node process already has this information because it must track it for other reasons.  Furthermore, this would require passing authentication information into the CNI plugin, including any certificates required and the apiserver IP address/hostname.

> One question I was surprised by - why does the CNI interface not pass enough information down to the network plugin to make a decision there?

CNI can only pass information that kubelet has; that will soon be extended to pass the PodSpec via https://github.com/containernetworking/cni/pull/247.  But that doesn't help openshift-sdn specific stuff like the HostSubnet and the VNID, which still need to be read out-of-band from CNI.

dcbw:
> Do you guys have a design doc / writeup / proposal to cover this work?

Yes, I'll mail you a link.

plugins/osdn/podinfoserver.go
smarterclayton: Talked this over with Jordan and David.  I think from a security perspective I can live with an API and a unix domain socket.  Some things that need to change / happen in order to merge it:

1. The handlers and structs sent back and forth need to be more strictly versioned and handled.  I.e., the path needs to be prefixed with a version `/v1/` and the objects should be in a versioned API package `sdn/localserver/api/v1` package and well documented
2. I want to nuke the watches from the SDN server component - it should use a kube client that is passed in (by the openshift node process) to make calls to the server, and potentially add an LRU cache.
3. I would recommend factoring your plugin as an http.Handler that is going to be instantiated by the node code (not directly used from inside openshift-sdn) for now
4. Use the existing Kube client instead of inventing your own - that handles rate limiting, etc
5. We'll have to sort through the unix domain socket and how it interacts with packaging, but at least for step 1 we are still going to run openshift node with both kubelet and SDN plugin (turned on) - the node process will boot the server, and the expectation the SDN sub binary will run inside of the node container.

plugins/osdn/bin/openshift-sdn-docker-setup.sh
dcbw: It gets done by the CNI plugin now:

```go
	err = ns.WithNetNSPath(args.Netns, func(hostNS ns.NetNS) error {
		hostVeth, contVeth, err = ip.SetupVeth(interfaceName, int(netInfo.MTU), hostNS)
		if err != nil {
			return fmt.Errorf("Failed to create container veth: %v", err)
		}
```

dcbw:
@smarterclayton after prototyping using the kube/openshift clients in the CNI plugin and discussing with @danwinship, I've decided to drop the podinfo server for now.  We may need it in the future if things become more complicated or contacting the apiserver becomes a performance issue.

I agree with your points (2) and (4) in https://github.com/openshift/openshift-sdn/pull/320#discussion_r68142557; the other points are moot with the latest commits.  We'll work on making those changes.

In the mean time, PTAL and let me know what you think.  Thanks!

dcbw:
@smarterclayton any thoughts on this approach?  One issue with not having communication between the CNI plugin and the openshift node process is that it somewhat complicates hostport/containerport redirect functionality (eg, like docker's "-p <host:container>" functionality) since the long-lived node process doesn't know when pods start/stop since the CNI plugin is now using its own k/os clients.

dcbw:
@liggitt any thoughts on the podinfo server vs. hitting up the apiserver directly here?

dcbw:
@eparis any thoughts here either?

dcbw:
Relevant comments are buried, so here they are:

https://github.com/openshift/openshift-sdn/pull/320#discussion_r67255635
https://github.com/openshift/openshift-sdn/pull/320#discussion_r68142557
https://github.com/openshift/openshift-sdn/pull/320#issuecomment-228480623

smarterclayton:
So I like the general approach of having CNI do more.  We may need to consider the perf impact, but it makes it truly modular rather than baked in.  Since you have the node credentials it's no worse.  If the hostPort stuff was rolled in (and it doesn't seem like there's a reason it can't be?) then it would remove the need for the node to do anything except maybe in the future be a caching proxy for the SDN tool.

dcbw:
> Why isn't hostPort managed by cni setup and tear down?

There currently isn't any way to get two things from the CNI plugin: (a) whether it handles HostPort itself, and (b) what interface name to do the iptables SNAT/DNAT on.

We're working on some of this upstream in CNI (https://github.com/containernetworking/cni/pull/145 and https://github.com/containernetworking/cni/pull/255) and will use this functionality in the Kubernetes CNI plugin then too.

dcbw:
> So I like the general approach of having CNI do more. We may need to consider the perf impact, but it makes it truly modular rather than baked in. Since you have the node credentials it's no worse. If the hostPort stuff was rolled in (and it doesn't seem like there's a reason it can't be?) then it would remove the need for the node to do anything except maybe in the future be a caching proxy for the SDN tool.

The HostPort is the only thing that's really icky about this right now.  I'm working to get that rolled into the kubernetes CNI network plugin but it's taking more upstream work to update the CNI spec than I thought.  But I expect that to be solved in the next month or two.

dcbw:
openshift-sdn is now merged into origin, so see https://github.com/openshift/origin/issues/9981 for the continuation of this PR.

PR #323, opened 2016-09-13
debug.sh will report incorrect message

pecameron:
When the /etc/hosts has extra spaces in the end of the line or the
node name is not the last token in the line the node is not discovered
and the result is error messages in the report.

This is fixed by using "getent ahostsv4" to look for the host name.
    
Fixes: 1328009

Signed-off-by: Phil Cameron <pcameron@redhat.com>

### DIFF
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -114,18 +114,10 @@ do_master () {
 	    continue
 	fi
 
-	if ! [[ $node =~ ^[0-9.]*$ ]]; then
-	    resolv_ip=$(awk '/\s'$node'$/ { print $1; exit; }' /etc/hosts)
-	    if [ -z "$resolv_ip" ]; then
-		resolv_ip=$(host $node 2>/dev/null | sed -ne 's/.*has address //p' | head -1)
-		if [ -z "$resolv_ip" ]; then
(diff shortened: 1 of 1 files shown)


pecameron:
@knobunc PTAL

hack/debug.sh
```diff
@@ -114,18 +114,10 @@ do_master () {
...
 
-	if ! [[ $node =~ ^[0-9.]*$ ]]; then
-	    resolv_ip=$(awk '/\s'$node'$/ { print $1; exit; }' /etc/hosts)
-	    if [ -z "$resolv_ip" ]; then
-		resolv_ip=$(host $node 2>/dev/null | sed -ne 's/.*has address //p' | head -1)
-		if [ -z "$resolv_ip" ]; then
-		    echo "Node $node: no IP address in either DNS or /etc/hosts"
-		fi
-	    fi
+	resolv_ip=$(getent ahostsv4 $node | grep STREAM | awk '{ print $1}')
```
danwinship: can simplify to just `| awk '/STREAM/ { print $1 }'`

this will return junk if the host has multiple IP addresses, but the old code would too... but you could change `print $1` to `print $1; exit` to fix that

pecameron:
@danwinship Thanks for the comment. I edited the file, PTAL

PR #324, opened 2016-09-27
IP Failover diagnostic script

pecameron:
As an OpenShift administrator, I want to know when my ipfailover is
causing my network problems.

Extend the nework debug script so that it can determine why keepalived
is not serving an address appropriately. Ideally we work out which
machine the address should be assigned to, then check that only that
machine has the address assigned, and (ideally) ping the address from
other machines to ensure connectivity.

Trello card:
https://trello.com/c/yFco7WvY/268-5-ip-failover-diagnostic-script-demo-supportability-p3-5

Signed-off-by: Phil Cameron pcameron@redhat.com

### DIFF
--- a/hack/ipf-debug.sh
+++ b/hack/ipf-debug.sh
@@ -1,0 +1,162 @@
+#!/bin/bash
+#set -o errexit
+set -o nounset
+set -o pipefail
+#set -x
+
+# This scripts finds the ipf deployments in the cluster and
+# analyzes the configuration
+
+# get all of nodes (some subset may be in use by ipf delpoyments)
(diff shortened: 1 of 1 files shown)


pecameron:
@knobunc Here is the ipf debug script. I made a pass through based on teh style guide. PTAL

knobunc:
LGTM [merge]

knobunc:
@pecameron can you add a bit to the troubleshooting guide -- https://docs.openshift.com/enterprise/3.1/admin_guide/sdn_troubleshooting.html

pecameron:
pr 3402 PTAL

PR #329, opened 2017-06-22
Correct the code that pulls the docker envfile

knobunc:
With the change to CNI the loop that pulled the envfile was consuming all of the other nodes... so the loop ended.  This checks that a file is present before calling cat.


knobunc:
@openshift/networking PTAL

PR #330, opened 2017-10-12
Add usage information to the debug script

imcsk8:
The debug script has some options that were not shown
in the usage information.

### DIFF
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -545,13 +545,17 @@ case "$1" in
 	    echo "  [from master]"
 	    echo "    $0"
 	    echo "  Gathers data on the master and then connects to each node via ssh"
+	    echo "    $0 --master"
+	    echo "  Runs only master debugging on local machine"
 	    echo ""
 	    echo "  [from any other machine]"
 	    echo "    $0 MASTER-NAME"
 	    echo "  Connects to MASTER-NAME via ssh and then connects to each node via ssh"
+	    echo "    $0 --node"
+	    echo "  Runs only node debugging on local machine"
 	    echo ""
-	    echo "  The machine you run from must be able to ssh to each other machine"
-	    echo "  via ssh with no password."
+	    echo "  In case of remote usage, the machine you run from must be able to ssh" 
+	    echo "  to each other machine via ssh with no password."
 	    exit 1
 	fi
 	;;
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -542,20 +542,29 @@ case "$1" in
     "")
 	if [ -z "$aos_master_service" ]; then
 	    echo "Usage:"
-	    echo "  [from master]"
+	    echo "  [From master]"
+	    echo "  Automatic:"
 	    echo "    $0"
 	    echo "  Gathers data on the master and then connects to each node via ssh"
+	    echo ""
+	    echo "  Manual:"
 	    echo "    $0 --master"
 	    echo "  Runs only master debugging on local machine"
 	    echo ""
-	    echo "  [from any other machine]"
+	    echo "  [From any node]"
+	    echo "  Automatic:"
 	    echo "    $0 MASTER-NAME"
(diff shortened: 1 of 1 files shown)


hack/debug.sh
```diff
@@ -545,13 +545,17 @@ case "$1" in
...
 	    echo "    $0"
 	    echo "  Gathers data on the master and then connects to each node via ssh"
+	    echo "    $0 --master"
+	    echo "  Runs only master debugging on local machine"
 	    echo ""
 	    echo "  [from any other machine]"
 	    echo "    $0 MASTER-NAME"
 	    echo "  Connects to MASTER-NAME via ssh and then connects to each node via ssh"
+	    echo "    $0 --node"
+	    echo "  Runs only node debugging on local machine"
```
danwinship: This requires some setup to have happened doesn't it? Like, you need to have the meta/ directory from a --master run?

If we're going to document the "manual" usage, it should be separate from the existing documentation of the "automatic" usage

imcsk8: you're right, but this is just the script usage, i guess a more detailed information should be in the official documentation or in the README file

danwinship: "From any other machine" there doesn't mean "from a node". It means "from a random machine outside the cluster". So it doesn't make sense to document `--master` up above but `--node` here.

The existing documentation isn't just enumerating command-line options, it's explaining the two ways you can run the script: you can run it on the master, or you can run it from any machine that can ssh to the master and the nodes. What you're documenting is a third way of running it, but you're mixing the documentation of that mode in with the documentation of the other two modes.

danwinship: > i guess a more detailed information should be in the official documentation

basically, this *is* the official documentation

hack/debug.sh
```diff
@@ -542,16 +542,29 @@ case "$1" in
...
 	    echo "  Gathers data on the master and then connects to each node via ssh"
 	    echo ""
-	    echo "  [from any other machine]"
+	    echo "  Manual:"
+	    echo "    $0 --master"
+	    echo "  Runs only master debugging on local machine"
+	    echo ""
+	    echo "  [From any node]"
+	    echo "  Automatic:"
 	    echo "    $0 MASTER-NAME"
```
danwinship: It's not "from any node", it's from literally *any* machine other than the master. Actually, it's almost certainly *not* going to be a node, since if your nodes can ssh to the master then the master can almost certainly ssh to the nodes as well, and so you might as well have just run it from the master in the first place.

This is for when the master doesn't have ssh keys to let it connect to nodes, and so instead, you run it from your desktop machine, where you do have an ssh key that lets you log in to both the master and the nodes.

hack/debug.sh
```diff
@@ -542,16 +542,29 @@ case "$1" in
...
-	    echo "  via ssh with no password."
+	    echo "  Manual:"
+	    echo "    $0 --node"
+	    echo "  Runs only node debugging on local machine"
+	    echo ""
+	    echo "  In case of Automatic usage, the machine you run from must be able to ssh" 
+	    echo "  to each other machine via ssh with no password."
+	    echo ""
+	    echo "  In Manual usage you have to login manually to the machine, upload this script," 
+	    echo " and create the meta directory."
```
danwinship: It's more than that. You need to copy the contents of the meta directory from the `--master` run to each node before running the script on the node, and you have to run the script *from* the meta directory. There might be some other constraints I'm not thinking of too... the nodes might expect the `node/` subdir to already exist, etc. It's not simple to do (or to describe how to do), because we don't really expect people to do it by hand. Which is why it wasn't documented...

What is this PR for? Why do you feel that we need to document these options? Especially since this script is slowly being deprecated in favor of `oadm diagnostics NetworkCheck` anyway...

imcsk8: it's related to this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1448447

PR #331, opened 2017-10-14
debug.sh: replace ping to curl

nak3:
Since some cloud services (especially EC2) disables ICMP by default,
debug.sh outputs the ping failure. Although users can workaround it by
setting security group, they would like to avoid such a workaround.

This patch changes node connectivity check to use curl instead of ping.

### DIFF
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -80,6 +80,7 @@ log_system () {
 }
 
 do_master () {
+    kubelet_port=10250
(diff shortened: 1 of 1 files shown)


hack/debug.sh
```diff
@@ -80,6 +80,7 @@ log_system () {
 }
 
 do_master () {
+    kublet_port=10250
```
danwinship: typo "kubelet"

(otherwise, sure, patch looks fine)

nak3:
Thank you. Fixed the typo.

PR #332, opened 2017-10-14
debug.sh: add since -3days to journalctl

nak3:
Since debug.sh collects journal log from all of the Nodes on the
cluster, if each cluster had xxx GB log, the collector host's
diskspace is overflowed. On top of that, current script takes much
time due to the collection time and scp of journal log.

This patch adds `--since -3days` option to journalctl, then collects
logs since 3days ago.

### DIFF
--- a/hack/debug.sh
+++ b/hack/debug.sh
@@ -47,7 +47,7 @@ log_service () {
     service=$2
     start_args=$3
 
-    echo_and_eval  journalctl -u $service                         &> $logpath/journal-$service
+    echo_and_eval  journalctl -u $service --since -3days          &> $logpath/journal-$service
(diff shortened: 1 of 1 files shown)


danwinship:
The unboundedness was intentional. In particular, we want the logs from the most recent service startup, if possible. (But also, if the service has been restarted recently, we want to see the logs from before the restart too.)

Maybe in some contexts we don't care about that as much, but if so, and if there were problems in some particular customer case, then we could provide a modified debug.sh for them to run. We don't expect people to run debug.sh other than when we have asked them to (since it requires an engineer to actually go through the output to look for problems, rather than diagnosing them itself).

PR #333, opened 2017-10-27
debug.sh: Use metadata.name instead of externalID to get node name

nak3:
When debug.sh gets Node name, it acquires `spec.externalID` from each
Node rather than `metadata.name`. Due to this, next command to get
IP fails, if `spec.externalID` and `metadata.name` are different.


danwinship:
debug.sh is no longer maintained

PR #334, opened 2017-10-30
Update SDN document reference

ganhuang:


danwinship:
ok but really we should replace this README entirely...

